PR #3, opened 2023-08-09
Initial dataset structure

wiederm:
## Description
Outline the basic dataset structure

This PR contains the structure for a `BaseDataset` that inherits from `torch.utils.data.Dataset` and contains functions to retrieve, manipulate and cache the downloaded dataset.
It also contains an implementation of a `QM9Dataset` that  provides functions that control from which remote location the dataset is retrieved and the transformations to generate an opimized npy file. It also implements the `__getitem__` and `__len__` calls necessary for the   `torch.utils.data.Dataset` class.

## Todos
  - [x] outline package structure
  - [x] define dataset module
 
## Status
- [x] Ready to go

### DIFF
--- a/modelforge/dataset/qm9.py
+++ b/modelforge/dataset/qm9.py
@@ -1,12 +1,12 @@
-from collections import defaultdict
-from typing import Any, Dict, Generator, List, Tuple
+import os
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Tuple
 
 import gdown
 import h5py
 import numpy as np
 import torch
 from loguru import logger
-
 from .dataset import BaseDataset
 
 
@@ -79,7 +79,12 @@ class QM9Dataset(BaseDataset):
     Also allows for lazy loading of data or caching in memory for faster operations.
     """
 
-    def __init__(self, dataset_name: str = "QM9", load_in_memory: bool = True) -> None:
+    def __init__(
+        self,
+        dataset_name: str = "QM9",
+        load_in_memory: bool = True,
+        test_data: bool = False,
+    ) -> None:
         """
         Initialize the QM9Dataset class.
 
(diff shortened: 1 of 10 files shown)


modelforge/tests/test_dataset.py
```diff
@@ -1,17 +1,116 @@
...
+    assert np.array_equal(padded_data[-1][-1], np.array([-1, -1, -1]))
+
+
 @pytest.mark.parametrize("dataset", [QM9Dataset])
 def test_download_dataset(dataset):
-    print(dataset.name)
-    dataset("tmp.hdf5")
+    """Test the downloading and batching of the dataset."""
+    d = dataset(dataset_name="tmp.hdf5", test_data=True)
+    print(d.dataset_name)
```
chrisiacovella, line 98: I'm guessing this print statement was leftover from testing.  Should this instead be an assert statement with the expected name?

modelforge/dataset/qm9.py
```diff
@@ -0,0 +1,204 @@
...
+import gdown
+import h5py
+import numpy as np
+import torch
+from loguru import logger
+from .dataset import BaseDataset
+
+
+class PlainQM9Dataset(torch.utils.data.Dataset):
+    """
```
chrisiacovella, line 13: What is the purposed of this class compared to !m9Dataset?

modelforge/dataset/qm9.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        Initialize the QM9Dataset class.
+
+        Parameters:
+        -----------
+        dataset_name : str
+            Name of the dataset, default is "QM9".
+        load_in_memory : bool
+            Flag to determine if the dataset should be loaded into memory, default is True.
+        test_data : bool
```
chrisiacovella, line 95: I was slightly confused by this at first.  Might want to just mark this something like "for development purposes."

chrisiacovella (approved):

modelforge/dataset/qm9.py
wiederm: I will remove this class.

modelforge/dataset/qm9.py
wiederm: agreed. changed the variable name.

modelforge/tests/test_dataset.py
wiederm: thanks! added the assert statment.

PR #5, opened 2023-08-25
Initial Base Classes and base functionality for Neural Network Potentials

wiederm:
## Description

This PR provides an NNP base class, NNPBase, to provide a flexible and extensible foundation for implementing various neural network potentials.
It also introduces an implementation of the SchNet architecture for predicting quantum mechanical molecular properties.

## Changes:
- Adds Schnet class in schnet.py for implementing the SchNet architecture.
- Introduces a BaseNNP class in models.py as a base class for neural network potentials.
- Includes various utility functions in utils.py like scatter_add, GaussianRBF, Dense, etc.
- Provides test coverage for all the new functionalities.
## Key Files:
- schnet.py
- models.py
- utils.py


yuanqing-wang:
I personally think that this parent class is a slight overkill.

- We don't need to handle derivatives explicitly, everything will be autograd and should happen outside of the module.
- I don't imagine we would need explicit output collection either.

For the parent class I think it should serve the follow purposes:

- enforce the shape of geometry
- enforce the shape of identity embedding
- allow other universal checks to be implemented

yuanqing-wang:
@wiederm I'm still a bit confused about the questions I've asked above---could you justify the necessity of those design choices a bit more? Thanks!

PR #11, opened 2023-09-14
dataset curation update

chrisiacovella:
## Description
Changes to hdf5 file formats. 

For efficiency in reading/writing, conformers will remain grouped, e.g., geometry will be an  m x n x 3 array, where m is number of conformations.  Parsing will be done when we read the curated datafile.    Data is tagged with an attribute denoting if it is series or not to easy in the parsing (allows routines to be more general for reading). 

Curation also switched to have a base class, to standardized the input and save some effort on unit tests.

I still need to update base hdf5 reader in the dataset class to include checking for NAN.  (need to merge code into module).  The desired behavior will be to simply not add a conformer if any of the desired properties are NAN.  

## Todos
  - [x] updated qm9 dataset reader
  - [x] set up basic ani base class
  - [x] bug in unzipping file: related to accessing data outside of the streamed chunk (do fill gzip first as it is no slower)
  - [x] address random slow downs (memory issue? need to try on machine with more memory) : Note: associated with indexing into the hdf5 file. creating an in-memory copy avoids this issue.
  - [x] add spice curation ( zenodo hdf5 file )

## Status
- [x] Ready to go


chrisiacovella:
Weird Mamba issue on Mac OS causing failures.   Not sure what is going on.  CI successful on linux.

chrisiacovella:
I think, once reviewed, we can merge this.  the qcarchive spice dataset will be a separate PR.  As I noted, this changes the structure of the hdf5 datasets a little bit (and the process for loading of them).  So it would be good to get this part finalized, so we don't have to retrain a lot of models.

modelforge/curation/qm9_curation.py
```diff
@@ -224,124 +232,143 @@ def _parse_xyzfile(self, file_name: str) -> dict:
...
-            data["smiles_gdb-17"] = smiles[0]
-            data["smiles_b3lyp"] = smiles[1]
-            data["inchi_Corina"] = InChI.split("\n")[0].split()[0].replace("InChI=", "")
-            data["inchi_B3LYP"] = InChI.split("\n")[0].split()[1].replace("InChI=", "")
-            data["geometry"] = np.array(geometry) * unit.angstrom
+            data_temp = {}
+            data_temp["name"] = file_name.split("/")[-1].split(".")[0]
+            data_temp["n_configs"] = 1
+            data_temp["smiles_gdb-17"] = smiles[0]
+            data_temp["smiles_b3lyp"] = smiles[1]
```
wiederm, line 278: is there a difference between the smiles strings?

modelforge/curation/qm9_curation.py
```diff
@@ -373,16 +400,25 @@ def process(
             output_path=self.local_cache_dir,
             force_download=force_download,
         )
```
wiederm, line 402: is this function imported?

modelforge/curation/qm9_curation.py
```diff
@@ -373,16 +400,25 @@ def process(
...
+        # clear out the data array before we process
+        self._clear_data()
```
wiederm, line 404: I am not complelty sure why this is necessary --- what is the motivation behind this call?

modelforge/curation/spice_curation.py
```diff
@@ -0,0 +1,256 @@
+from modelforge.utils.units import *
+
+from modelforge.curation.curation_baseclass import *
+
+
+class SPICE_1_1_4_curation(dataset_curation):
```
wiederm, line 6: Is it possible to define a SpiceCuration(DatasetCuration) that takes as keyword argument the specific version we want to pull down?

wiederm (approved):
Great work! This is awesome!

I left a few non-blocking comments mostly about inconsistent Python style with the current code. I am open to discussing this and adopting it, my motivation is to keep consistencu between modules.

modelforge/curation/ani1x_curation.py
```diff
@@ -0,0 +1,304 @@
+from modelforge.utils.units import *
+
+from modelforge.curation.curation_baseclass import *
```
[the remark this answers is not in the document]
chrisiacovella: this is one I'm not exactly sure how to handle, since I'm not really trying to import a function or a class, but rather code that adds an additional context to open units to be able to to do transformations between, e.g., energy and energy/mol . 

```
c = unit.Context("chem")
c.add_transformation(
    "[force] * [length]",
    "[force] * [length]/[substance]",
    lambda unit, x: x * unit.avogadro_constant,
)
c.add_transformation(
    "[force] * [length]/[substance]",
    "[force] * [length]",
    lambda unit, x: x / unit.avogadro_constant,
)
c.add_transformation(
    "[force] * [length]/[length]",
    "[force] * [length]/[substance]/[length]",
    lambda unit, x: x * unit.avogadro_constant,
)
c.add_transformation(
    "[force] * [length]/[substance]/[length]",
    "[force] * [length]/[length]",
    lambda unit, x: x / unit.avogadro_constant,
)

c.add_transformation(
    "[force] * [length]/[length]/[length]",
    "[force] * [length]/[substance]/[length]/[length]",
    lambda unit, x: x * unit.avogadro_constant,
)
c.add_transformation(
    "[force] * [length]/[substance]/[length]/[length]",
    "[force] * [length]/[length]/[length]",
    lambda unit, x: x / unit.avogadro_constant,
)


unit.add_context(c)

```

modelforge/curation/qm9_curation.py
chrisiacovella: Yes, they can be; calculated using two different approaches I believe (which may have different assumptions that identify connectivity).

modelforge/curation/curation_baseclass.py
```diff
@@ -0,0 +1,192 @@
...
+        output_file_dir: str, optional, default='./'
+            Location to write the output hdf5 file.
+        local_cache_dir: str, optional, default='./qm9_datafiles'
+            Location to save downloaded dataset.
+        convert_units: bool, optional, default=True
+            Convert from e.g., source units [angstrom, hartree]
+            to output units [nanometer, kJ/mol]
+        """
+
+        from modelforge.utils.misc import mkdir
```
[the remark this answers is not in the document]
chrisiacovella: That is essentially what is being called by the mkdir function I added...I had a "good" reason for doing it this way initially, but I'm not exactly sure what that is anymore (I think it was related to testing of creating temporary directories, but that need has been removed).  So yeah, the specialization I was going for is not being used at this point so I'll just remove that function.

modelforge/curation/qm9_curation.py
chrisiacovella: Good catch.  it was not (and this part of the code was not being checked by CI because it takes too long).  All the more reason for the more thorough offline suite checking.

modelforge/curation/qm9_curation.py
chrisiacovella: The `self.data` list (which stores the processed records before being written to the hdf5 file)  is initialized in the constructor.  If someone were to initialize the class, and then call the `process()` function multiple types, it would just append the same records multiple times to `self.data`, if we didn't clear data in-between calls, which is just what `self._clear_data()` does.  

Since these aren't really functions that people are going to be using, really, directly at all, I didn't bother putting in any checking where if you call process a second time it would check the contents of data to see if only parts of it need to be reprocessed.   
wrapping data.clear() in it's own function was purely just for making improving the codecov metrics.

modelforge/curation/spice_curation.py
chrisiacovella: I think we can.  I'll have to check with Peter Eastman on what the specific differences are and whether that is actually something useful (I think e.g., 1.1.2 is just a subset of the 1.1.4 dataset).  It would be an easy chance to make in the future if that is ultimately relevant.  Our interest is mostly on the openff level theory for spice, that is only available from qcarchive at the moment.

PR #28, opened 2023-10-11
Fixed dim issue with SchNET implementation

wiederm:
## Description
THis PR fixes a bug in the padding of the embedding and interaction layer of SchNET.

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/schnet.py
+++ b/modelforge/potential/schnet.py
@@ -77,13 +77,18 @@ class Schnet(BaseNNP):
         """
         # initializing x^{l}_{0} as x^l)0 = aZ_i
         Z = inputs["Z"]
-        x = self.embedding(Z)  # shape (batch_size, n_atoms, n_atom_basis)
-        mask = Z == 0
+        mask = Z != 0
         pairlist = self.calculate_distances_and_pairlist(mask, inputs["R"])
+
+        # unpad Z
+        # Z = Z_[mask].flatten()
+        x = self.embedding(Z)  # shape (batch_size * n_atoms, n_atom_basis)
 
         representation = self.representation(
             pairlist
         )  # shape (batch_size, n_atoms, n_atom_basis)
+
+        # unpad
 
         for interaction in self.interactions:
             v = interaction(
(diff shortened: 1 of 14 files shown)


modelforge/tests/helper_functinos.py
chrisiacovella, line 1: typo in the name here "helper_functinos.py" should be "helper_functions.py"

modelforge/potential/schnet.py
```diff
@@ -237,7 +244,7 @@ def forward(self, pairlist: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
 
         idx_i, idx_j = atom_index12[0], atom_index12[1]
 
-        return x
+        return {"f_ij": f_ij, "idx_i": idx_i, "idx_j": idx_j, "rcut_ij": rcut_ij}
 
 
 class LighningSchnet(Schnet, LighningModuleMixin):
```
chrisiacovella, line 250: Two things here:

- We should probably consistent of `Schnet` vs `SchNet`
- minor typo in Lightning (forgot the t).

modelforge/potential/schnet.py
```diff
@@ -162,20 +180,15 @@ def forward(
...
         return x
 
 
 class SchNetRepresentation(nn.Module):
     def __init__(
         self,
-        n_atom_basis: int,
-        n_filters: int,
-        n_interactions: int,
         cutoff: float = 5.0,
```
chrisiacovella, line 192: It would be good here to specify the units (even without using the unit package, just annotate what units cutoff should be in).  Also I think the data we are loading in is nanometers, so this should be 0.5 (since 5.0 angstrom is used in schnet, right)?

chrisiacovella (changes requested):
A few minor things at this point.  I still need to go through more thoroughly after I fix up my install on my machine.

modelforge/potential/schnet.py
wiederm: that's great catch, thank you!

chrisiacovella:
I'll note all the failures are coming from the changes to zenoto that break some of the functions I wrote (see #31 ). I have these updated/removed in a separated PR (#30 ), so those failing tests can be ignored for now.

chrisiacovella (approved):
Looking through this I'm not seeing any obvious issues.  I think we can merge now, that way we can get to a point to do more detailed tests/comparisons.

Issue #29, opened 2023-10-12
HDF5 file structure; encoding information about array axis


chrisiacovella:
The general design of the curated hdf5 files includes grouping together conformers into a single entry.  For example, coordinates are stored as an [`m`, `n`, `3`]  array for each molecule (where `m` is number of conformers, `n` is number of atoms, `3` correspond to x,y,z).

For a quantity such as atomic_numbers, which will be fixed for all conformers, we do not need to make `m` copies of the same data.  To differentiate between these two types of data, I added in an attribute call "series" where True would indicate we have a series of per-conformer entries, and False would indicate something like atomic_numbers, which apply to all the conformers. 

-  @ArnNag found that I was not consistent with how this was defined for the qm9 (which did not have explicit conformers defined); I marked all values as series=False, which since we don't have conformers is valid, but it means that the underlying arrays will have different shapes than a dataset with conformers (which might make routines to handle this data less generalizable).  This code needs to be updated such that any quantity that could be per-conformer, is properly marked (and underlying data arrays have the same shape as if there were multiple conformers).  

- discussion of the aforementioned issue has led to a more in-depth conversation in terms of providing a quick way to determine what the data represents.  currently "series" will tell us if axis=0 represents different conformers.  We still do not know if the the underlying data represents a per-atom quantity or a per-molecule quantity.  If the value of a given entry for a conformer were a scalar value, that would certainly indicate a per-molecule quantity (e.g., energy), but not all per-molecule quantities need to be scalars (e.g., dipole moments).   As such, we should encode a second piece of information regarding per-molecule or per-atom.  We do not necessarily need a second attribute, but could change the series attribute to be something like `format` which a string that encodes the format:

{conformer_info}_{property_info}

conformer_info could be: `series` or `single`
proper_info could be: `atom` or `mol`

More generally, I think need to ensure that, e.g., when defining energy per molecule, we have an array of size [m,1] rather than [m], such that how we interrogate the array (i.e., look at .shape) is the same for a quantity that may have a spatial dependence (e.g., shape [m,3]).

PR #30, opened 2023-10-13
Spice openff

chrisiacovella:
## Description
This PR does two main things. 
 (1) it provides curation routines for the spice dataset at the openff level of theory  fetching from qcarchive
(2) it makes array shapes in the generated hdf5 files consistent (see Issue: #29 ) 
(3) This also addresses breaking changes to the zenodo api (see Issue #31 ), fixing routines we use and removing ones I wrote that we ultimately did not end up using. 

## Todos
  - [x] As discussed in issue #29 , need to modify the attribute for each field to denote if a property is per-atom or per-molecule, in addition to the "series" info (that says if the property contains information that varies per conformer).  This is necessary because simply looking at the shape of an array is in itself insufficient.  This should not be hard, but need to decide on the most appropriate way to define it (and annotate in the code).
  - [x] Add in checksum verification for figshare/zenodo downloader

## Questions
- [x] For Spice openff, do we want the hdf5 file to include the dispersion correction energy/forces separate or only have a single corrected value?  Is there any reason for these values to be separate? Consensus is to only include dispersion corrected energies and forces. 

## Status
- [x] Ready to go


chrisiacovella:
I think this is now ready to go.  Some of the data loading is a bit hacky, as some of the changes to the shape of the underlying data arrays was done purposefully to make it easier for the diagonal batching that is being worked on.  (e.g., I need to reshape the atomic numbers of shape [n,1], i.e., [[12],[1], [1], [1], [1]] to a [n] array);  the pad_to_max_length function was updated then to operate on np arrays rather than lists, since I think all of our underlying data should be of the form of an array.  Similarly, energy is stored is an [m,1] array.  For QM9, since m =1, this is just a single float value. To keep things compatible at this moment, this single float value is extracted and appended to the array (adding an array was failing; not sure how much effort to put into this considering we will be changing the approach).

chrisiacovella:
@ArnNag Can you take a look?

modelforge/dataset/utils.py
```diff
@@ -237,10 +237,10 @@ def pad_to_max_length(data: List[np.ndarray]) -> List[np.ndarray]:
     List[np.ndarray]
         List of padded arrays.
     """
-    max_length = max(len(arr) for arr in data)
+    max_length = max(arr.shape[0] for arr in data)
 
     return [
-        np.pad(arr, (0, max_length - len(arr)), "constant", constant_values=0)
```
wiederm, line 243: nice!

modelforge/tests/test_remote.py
```diff
@@ -5,8 +5,6 @@
 
 from modelforge.utils.remote import *
 
-IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
-
 
 @pytest.fixture(scope="session")
 def prep_temp_dir(tmp_path_factory):
```
wiederm, line 10: might make sense to put this in a `conftest.py` file

wiederm (approved):
This is great! I don't see any blocker

modelforge/tests/test_remote.py
chrisiacovella: Ah, so this just got left in accidentally, as I was testing some different ways to mark tests for local vs. CI.

PR #34, opened 2023-10-20
infrastructure work

wiederm:
This PR introduces the new API for NNPs that @chrisiacovella and I discussed with @jchodera .

The updated version contains an AbstractBaseNNP that mimics the previous BaseNNP class. It has two methods:
`forward` implemented as an abstract method providing hints for the preferred input signature `Dict[str, torch.Tensor]` and documentation, as well as `input_checks` that currently checks for the two keys that need to be present, `atomic_numbers` and `positions`, as well as for the dimensionality of the `torch.Tensor`.

The former `BaseNNP` is now a more specialized implementation for an NNP, which currently splits the calculation of the neighbor pairs (which is now in the `forward` method), and calls a NNP specific `_forward` method that will be responsible for perform the data representation, atom interaction and the calculation of the atom-wise contributions and accumulation in global prediction.

The input signature for the AbstractBaseNNP (and following NNPs) 

```python
# Abstract Base Class for every NNP
class AbstractBaseNNP(nn.Module, ABC):
    """
    Abstract base class for neural network potentials.

    Methods
    -------
    forward(inputs: Dict[str, torch.Tensor]) -> torch.Tensor
        Abstract method for neighbor list calculation and forward pass in neural network potentials.
    _forward(inputs: Dict[str, torch.Tensor]) -> torch.Tensor
       Abstract method for forward pass in NNP
    input_checks(inputs: Dict[str, torch.Tensor])
        Perform input checks to validate the input dictionary.
    """

    def __init__(self):
        """
        Initialize the AbstractBaseNNP class.
        """
        super().__init__()

    @abstractmethod
    def forward(self, inputs: Dict[str, torch.Tensor]) -> torch.Tensor:
        """
        Abstract method for neighbor list calculation and forward pass in neural network potentials.

        Parameters
        ----------
        inputs : Dict[str, torch.Tensor]
            - 'atomic_numbers', shape (nr_systems, nr_atoms), 0 indicates non-interacting atoms that will be masked
            - 'total_charge', shape (nr_systems, 1)
            - 'positions', shape (n_atoms, 3)
            - 'boxvectors', shape (3, 3)

        Returns
        -------
        torch.Tensor
            Calculated output; shape is implementation-dependent.

        """
        pass

    @abstractmethod
    def _forward(self, inputs: Dict[str, torch.Tensor]):
        """
        Abstract method for forward pass in neural network potentials.
        This method is called by `forward`.

        Parameters
        ----------
        inputs: Dict[str, torch.Tensor]
            - 'pairlist': Dict[str, torch.Tensor], contains:
                - pairlist, shape (n_paris,2)
                - r_ij, shape (n_pairs, 1)
                - d_ij, shape (n_pairs, 3)
                - 'atomic_subsystem_index' (optional), shape n_atoms

        """
        pass
```


modelforge/dataset/dataset.py
```diff
@@ -189,12 +202,12 @@ def _from_hdf5(self) -> None:
                                     not_nan = False
                                     break
                             else:
-                                #I don't like this, it is not general, but necessary at this point to
+                                # I don't like this, it is not general, but necessary at this point to
```
chrisiacovella, line 205: You could probably remove my comment :-)

modelforge/potential/models.py
```diff
@@ -29,54 +39,77 @@ def __init__(self, cutoff: float = 5.0):
...
 
         Returns
         -------
         torch.Tensor, shape [n_pairs, 3]
             Displacement vector between atom pairs.
         """
-        coordinates = R.flatten(0, 1)
-        selected_coordinates = coordinates.index_select(0, atom_index12.view(-1)).view(
+        coordinates = positions.flatten(0, 1)
+        selected_coordinates = coordinates.index_select(0, atom_pairs.view(-1)).view(
```
chrisiacovella, line 60: I don't think the logic here is correct. 

let us assume we have a positions tensor shape [2,2,3], so two two batches, each with 2 atoms (with one pair between the two atoms in each batch):

`
positions = torch.tensor([[[0. , 0. , 0. ],[1. , 0. , 0. ]],[[3. , 0. , 0. ],[3. , 0.5, 0. ]]])
atom_pairs = torch.tensor([[0, 1],[2, 3]])
`
If we apply the logic in the code, selected_coordinates ends up with:

`
tensor([[[0.0000, 0.0000, 0.0000],
         [1.0000, 0.0000, 0.0000]],

        [[3.0000, 0.0000, 0.0000],
         [3.0000, 0.5000, 0.0000]]])
`
so therefore the the difference (selected_coordinates[0]-selected_coordinates[1]) ends up:

`
tensor([[-3.0000,  0.0000,  0.0000],
        [-2.0000, -0.5000,  0.0000]])
`
Obviously the difference between atoms in two batches, not within the same batch. 

I think line 60 should be:

`
selected_coordinates = coordinates.index_select(0, pairs.view(-1)).view(2,-1).split(3,-1)
`

for this selected_coordinates are:

`
(tensor([[0., 0., 0.],
         [3., 0., 0.]]),
 tensor([[1.0000, 0.0000, 0.0000],
         [3.0000, 0.5000, 0.0000]]))
`

and thus the difference is what we'd expect from the input data.  

`
tensor([[-1.0000,  0.0000,  0.0000],
        [ 0.0000, -0.5000,  0.0000]])
`

modelforge/potential/utils.py
```diff
@@ -304,45 +303,32 @@ def _distance_to_radial_basis(
 
 # taken from torchani repository: https://github.com/aiqm/torchani
 def neighbor_pairs_nopbc(
-    mask: torch.Tensor, R: torch.Tensor, cutoff: float
+    padding_mask: torch.Tensor, coordinates: torch.Tensor, cutoff: float
```
chrisiacovella, line 306: We should be consistent, positions or coordinates in our function variable names.

modelforge/tests/test_painn.py
```diff
@@ -1,11 +1,89 @@
...
+        "./qm9.db",
+        batch_size=64,
+        num_train=1000,
+        num_val=1000,
+        transforms=[
+            trn.ASENeighborList(cutoff=5.0),
+            trn.RemoveOffsets(QM9.U0, remove_mean=True, remove_atomrefs=True),
+            trn.CastTo32(),
+        ],
+        property_units={QM9.U0: "eV"},
```
chrisiacovella, line 46: we aren't using eV as our units....same as above.  comment/annotate what is going on

modelforge/tests/test_painn.py
```diff
@@ -1,11 +1,89 @@
...
+    if not os.path.exists("qm9tut"):
+        os.makedirs(qm9tut)
+
+    qm9data = QM9(
+        "./qm9.db",
+        batch_size=64,
+        num_train=1000,
+        num_val=1000,
+        transforms=[
+            trn.ASENeighborList(cutoff=5.0),
```
chrisiacovella, line 42: we are using nm, I'm assuming this is angstroms which would be inconsistent.  We should make sure that we are commenting and annotating "This calls schnetpack which uses different units..., the unit system is...")

modelforge/tests/test_models.py
```diff
@@ -11,20 +10,35 @@
...
+#         dataset,
+#         mode="fit",
+#     )  # split_file="modelforge/tests/qm9tut/split.npz")
+#     output = initialized_model(inputs)
+#     print(output)
+#     assert output.shape[0] == 64
+#     assert output.shape[1] == 1
 
 
 def test_pairlist_simple_data():
```
chrisiacovella, line 44: If my comment is accurate about the pair list computation being off, these tests need to be rewritten because they aren't catching the error I saw.

chrisiacovella (changes requested):
I think there is an issue with the pair vector computation.

modelforge/potential/models.py
chrisiacovella: This is not incorrect; the issue stems from the order of indices in the atom_pairs.  It is NOT `[[a_i, a_j], [b_i, b_j]]`, but rather `[[a_i, b_i], [a_j, b_j]]` (where `a` is the first batch, `b` is the second batch, for pairs `i` and `j`).   It appears the docstring is not correct.  Should be shape [2, n_pairs], where atom_pairs[0] lists all of the index `i` atoms in the pair  , atom_pairs[1]  refers to the `j` .

chrisiacovella: Docstring has been updated so this has been resolved.

modelforge/dataset/dataset.py
wiederm: :-) removed

modelforge/potential/utils.py
wiederm: Thanks! Updated this!

chrisiacovella (approved):
just a few non-blocking comments about giving some better annotating in one of the tests.

modelforge/tests/test_painn.py
wiederm: I added a comment at the beginning of the test to point out that these units aren't yet consistent with our unit system. This might change in the future though.

PR #35, opened 2023-10-20
Added in reference energy calculation for qm9.  

chrisiacovella:
Reference energy (i.e., energy of the constituent atoms in vacuum) is computed for the QM9 dataset from the data made available in the paper SI; other thermochemical properties calculated and added to the datafile, reference energy at 298.15K, reference enthalpy at 298.15K and reference free energy at 298.15K.    This formation energy is also calculated and added to the data files for 0K (i.e., internal energy - reference energy).

equivalent shifted enthalpy, free energy, and internal energy at 298.15K could also be calculated, but given the very close similarity to internal energy at 0K (and the fact we will be training off internal energy at 0K), I do not think this will be necessary; of course, since the references are provided for each of these, they could be calculated by a user 

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/curation/qm9_curation.py
+++ b/modelforge/curation/qm9_curation.py
@@ -98,7 +98,23 @@ class QM9Curation(DatasetCuration):
                 "u_in": unit.hartree,
                 "u_out": unit.kilojoule_per_mole,
             },
+            "formation_energy_at_0K": {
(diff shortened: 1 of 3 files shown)


wiederm (approved):
This is great! :tada:

PR #37, opened 2023-10-31
Training schedule

wiederm:
## Description
Provide the initial training routines for NNPs

## Todos
  - [ ] pytorch-lighning training routines
  - [ ] Ray implementation of hyperparamter scan

## Status
- [ ] Ready to go


yuanqing-wang:
@wiederm if we choose to always stick to Lightening, we don't need this PR anymore, right? I can draft another PR with ray tune.

wiederm:
Yes, that's rigt.

PR #38, opened 2023-10-31
Flexible NNP implementations

wiederm:
## Description
PaiNN (and to a lesser degree SchNET) has some implicit assumptions about the number of layers and dimensions. This PR aims to expose the relevant parameters and make it customizable.

## Todos
  - [x] flexible EnergyReadout
  - [x] flexible Filters

## Status
- [x] Ready to go


modelforge/potential/__initi__.py
chrisiacovella, line 1: This filename has a typo (`initi`, rather than `init`)

modelforge/potential/pain.py
```diff
@@ -35,11 +34,11 @@ def __init__(
         Parameters
             ----------
             embedding : torch.Module
```
chrisiacovella, line 36: I think it would be good to improve the doc string here (i.e., give a bit more description about the expectation of the embedding), and also give an example below.

modelforge/potential/pain.py
chrisiacovella, line 1: I think it would be better to name this file painn.py (with 'nn' instead of a single 'n') for consistency.

chrisiacovella (approved):
Mostly just non-blocking comments.

modelforge/potential/pain.py
wiederm: agreed! sorry for that :-)

modelforge/potential/__initi__.py
wiederm: thanks!

modelforge/potential/schnet.py
```diff
@@ -105,19 +113,22 @@ def __init__(self, nr_atom_basis: int, nr_filters: int, nr_rbf: int = 20) -> Non
             Number of atom basis, defines the dimensionality of the output features.
         nr_filters : int
             Number of filters, defines the dimensionality of the intermediate features.
-        nr_rbf : int, optional
-            Number of radial basis functions. Default is 20.
+        nr_rbf : int
```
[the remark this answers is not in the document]
wiederm: added assert statments. these might also be useful in other places

modelforge/potential/pain.py
wiederm: added more informative docstrings

Issue #39, opened 2023-10-31
Split on molecules and not conformations in dataloader


wiederm:
As mentioned in the tile, we are currently splitting on conformations and should change this to a splitting on molecules

ArnNag:
I think I removed the record index @chrisiacovella was referring to since it was redundant with `dataset["n_confs"]`, which is the count of conformers in each record. So in order to get the split indices, the splitting strategy's split method would have to shuffle an index with the length of `dataset["n_confs"]` , split these indices, then convert this to `atomic_subsystem_indices_referencing_dataset` (as in https://github.com/choderalab/modelforge/blob/48281ac742d476d7e9371c59e9c7b586205c8dce/modelforge/dataset/dataset.py#L569).

wiederm:
Any progress here? Is there anything I can help with?

ArnNag:
#41 should resolve this.

chrisiacovella:
It seems to make sense to remove the record_index array (that was a bit of a kludge to work with the prior structure of the reader).  

I put together some example datafiles that basically just duplicate some of the data in the qm9 dataset (just duplicating some of the underlying arrays), that should be useful for testing (we can add these in as additional tests for the qm9 dataset, maybe change from a single bool "unit_testing" to some sort of specification of type of test we want).

10 records total, 24 total conformers:
https://drive.google.com/file/d/1xdGoqEkWxmxysJIDiienX4TOeZccCT9s/view?usp=sharing
100 records, total, 200 total conformers:
https://drive.google.com/file/d/1Ror0KnopG0KR4EvV8yagWq0G5MsU66Md/view?usp=sharing

PR #42, opened 2023-11-07
Shuffle by record rather than conformer 

ArnNag:
## Description
Address #39 

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/dataset/utils.py
+++ b/modelforge/dataset/utils.py
@@ -275,6 +275,9 @@ def random_record_split(
             indices.extend(dataset.get_series_mol_idxs(record_idx))
         indices_by_split.append(indices)
 
+    if sum([len(indices) for indices in indices_by_split]) != len(dataset):
+        raise ValueError("Sum of all split lengths does not equal the length of the input dataset!")
+
     return [Subset(dataset, indices_by_split[split_idx]) for split_idx in range(len(lengths))]
 
 
--- a/modelforge/dataset/utils.py
+++ b/modelforge/dataset/utils.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections import OrderedDict
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


modelforge/dataset/utils.py
```diff
@@ -126,6 +126,161 @@ def split(self, dataset: "TorchDataset") -> Tuple[Subset, Subset, Subset]:
...
+
+        # add 1 to all the lengths in round-robin fashion until the remainder is 0
+        for i in range(remainder):
+            idx_to_add_at = i % len(subset_lengths)
+            subset_lengths[idx_to_add_at] += 1
+
+        lengths = subset_lengths
+
+        for i, length in enumerate(lengths):
+            if length == 0:
```
wiederm, line 261: this results in an empty dataset, can this be intended behavior?

modelforge/tests/test_dataset.py
```diff
@@ -174,22 +231,22 @@ def test_dataset_generation(dataset):
...
+    train_dataset, val_dataset, test_dataset = RandomRecordSplittingStrategy().split(dataset)
 
     energy = train_dataset[0]["E_label"].item()
     assert np.isclose(energy, -412509.9375)
     print(energy)
 
     try:
-        RandomSplittingStrategy(split=[0.2, 0.1, 0.1])
+        RandomRecordSplittingStrategy(split=[0.2, 0.1, 0.1])
     except AssertionError:
```
wiederm, line 245: that should be `except AssertionError as e`

wiederm (approved):
Thank you! This looks really good!

modelforge/tests/test_dataset.py
ArnNag: sorry, used the online editor since i was working on another commit locally and didn't catch this :(

modelforge/dataset/utils.py
ArnNag: Perhaps the user intends to create an empty dataset, but this will warn the user that they are doing so. I copied this from the implementation of `torch.utils.data.random_split`

wiederm (approved):

modelforge/dataset/utils.py
wiederm: Thanks for clarifying

PR #48, opened 2023-11-13
Keep input dimensions consisten

wiederm:
## Description
Input dimensions for `inputs["atomic_numbers"]` has shape (nr_of_atoms_in_batch, 1), which was flattended in the `BaseNNP` before. Now we keep dimensions and let the NNP decide what to do with it.

## Question
Should we change the name of the input property `atomic_number` 
https://github.com/choderalab/modelforge/blob/6dab8a1a376a8b0cf83d84ac927793154239e272/modelforge/potential/models.py#L172
 to `atomic_properties`? The way we are approaching the embedding in this PR makes it flexible to have other properties defined in this tensor with the now flexible shape `(nr_of_atoms_in_batch, *, *)`. What do you think @ArnNag ?

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -6,25 +6,66 @@ import torch.nn as nn
 
 
 class SlicedEmbedding(nn.Module):
-    def __init__(
-        self, numbers_of_atoms_in_batch: int, embedding_dim: int, sliced_dim: int = 1
-    ):
+    """
+    A module that performs embedding on a selected slice of input tensor.
+
+    Parameters
+    ----------
+    max_Z : int
+        Highest atomic number to embed, this will define the upper bound of the vocabulary that is used for embedding.
+    embedding_dim : int
+        Size of the embedding.
+    sliced_dim : int, optional
+        The dimension along which to slice the input tensor (default is 0).
+        This is relevant since the input dimensions are (n_atoms_in_batch, nr_of_properties, property_size), but we want to embed a specific
+        property.
+
+    Attributes
+    ----------
+    embedding : nn.Embedding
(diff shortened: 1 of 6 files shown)


modelforge/potential/utils.py
```diff
@@ -5,6 +5,48 @@
 import torch.nn as nn
 
 
+class SlicedEmbedding(nn.Module):
+    def __init__(
```
chrisiacovella, line 9: Can you add in a docstring for this class?

modelforge/tests/helper_functions.py
```diff
@@ -38,8 +38,10 @@ def setup_simple_model(
         Initialized model.
     """
     from modelforge.potential import CosineCutoff, GaussianRBF
+    from modelforge.potential.utils import SlicedEmbedding
 
-    embedding = torch.nn.Embedding(max_atomic_number, nr_atom_basis)
+    embedding = SlicedEmbedding(max_atomic_number, nr_atom_basis, sliced_dim=0)
+    assert embedding.embedding_dim == nr_atom_basis
```
chrisiacovella, line 44: There doesn't seem to really be a test of the functionality of the `SlicedEmbedding` class.

chrisiacovella (changes requested):
Not too much to do on this, but I think there needs to be some tests of `SlicedEmbedding` class.  right now it is only checking the dimension that gets set in the class.

modelforge/potential/utils.py
wiederm: sure!

chrisiacovella (approved):
Looks good, lets get this merged!

PR #55, opened 2024-01-30
Seperating Neighborlist and Pairlist

wiederm:
## Description
Our current Pairlist included a possible cutoff for neighbor calculations; this has been refactored in two separate classes to make sure we use the intended calculation. Typically, with only a few use cases, we want the Pairlist for training. 

## Todos
  - [x] Refactor current PairList in a separate class for pairlist and neighborlist calculation, one taking a cutoff argument

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/models.py
+++ b/modelforge/potential/models.py
@@ -97,6 +97,58 @@ class PairList(nn.Module):
         }
 
 
+class NeighbourList(PairList):
+    def __init__(self, cutoff: float, only_unique_pairs: bool = False):
+        """
+        Initialize PairList.
+
+        Parameters
+        ----------
+        cutoff : float
+            Cutoff distance for neighbor calculations.
+        only_unique_pairs : bool, optional
+            If set to True, only unique pairs of atoms are considered, default is False.
+        """
+        super().__init__(only_unique_pairs=only_unique_pairs)
+        from .utils import neighbor_list_with_cutoff
+
+        self.calculate_pairs = neighbor_list_with_cutoff
+        self.cutoff = cutoff
+
+    def forward(
+        self, positions: torch.Tensor, atomic_subsystem_indices: torch.Tensor
+    ) -> Dict[str, torch.Tensor]:
+        """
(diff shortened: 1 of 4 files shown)


modelforge/potential/models.py
```diff
@@ -90,10 +84,60 @@ def forward(
             - 'd_ij' : torch.Tenso, shape (3, n_pairs)
```
chrisiacovella, line 84: minor typo: "torch.Tensor" I also think you swapped the shape of r_ij and d_ij in the doc-string

modelforge/potential/models.py
```diff
@@ -90,10 +84,60 @@ def forward(
...
+            "pair_indices": pair_indices,
+            "d_ij": self.calculate_d_ij(r_ij),
+            "r_ij": r_ij,
+        }
+
+
+class NeighbourList(PairList):
+    def __init__(self, cutoff: float, only_unique_pairs: bool = False):
+        """
+        Initialize PairList.
```
chrisiacovella, line 103: Might be good to update the docstring here.

modelforge/potential/models.py
```diff
@@ -90,10 +84,60 @@ def forward(
...
+        r_ij = self.calculate_r_ij(pair_indices, positions)
+
+        return {
+            "pair_indices": pair_indices,
+            "d_ij": self.calculate_d_ij(r_ij),
+            "r_ij": r_ij,
+        }
+
+
+class NeighbourList(PairList):
```
chrisiacovella, line 100: I think it would be good to use one version of neighbor/neighbour.  Most of the other places you use the non-british version of neighbor.

chrisiacovella (approved):
A few very minor changes.

modelforge/potential/models.py
wiederm: thanks for catching that!

modelforge/potential/models.py
wiederm: Yes! Neighbor it is

Issue #54, opened 2024-01-30
Make sure units are handled consistently and are aligned with the openmm unit system.


wiederm:
_Originally posted by @chrisiacovella in https://github.com/choderalab/modelforge/pull/49#discussion_r1471564496_

wiederm:
Closed by PR #56

Issue #32, opened 2023-10-19
Adding unit checking/conversion to entry points


chrisiacovella:
As discussed in our roadmap, we want to have units associated with entry points.   

Two key spots in the code that require this: 
- loading hdf5 files
 - The use of the curated datasets should already ensure proper units, but checking the units against expected ones would provide an additional consistency check.  
- The expected units should be defined in the context of the 'available_properties' for a given dataset (i.e. convert the list of properties to a dict). 
  -  If checking units is prohibitively slow for all records, we can have the default behavior be to simply check the first entry (full checking can be enabled by an optional flag)
  -  We should print to the log and screen the specific units that are being used for clarity. 
  - Since we are using curated datasets, I do not think we need to implement unit conversions (although this is would be trivial); this is mostly to ensure we do not have inconsistencies 
  
- passing cutoffs (e.g., distance cutoffs, force cutoffs):
   -  Rather than just encoding the expected unit in the doctoring, we should require a user to assign the units to the variable as an openff-unit quantity (e.g., 0.5*unit.nanometer). 
   - In this case I think we should absolutely support conversion to the appropriate unit (with a warning/note to the user that we are converting to the expected unit)

wiederm:
In my opinion, this has been addressed in the linked PRs. Closing for now.

PR #56, opened 2024-01-30
Adding unit checks to the public API

wiederm:
## Description
Currently, we are not enforcing units at several public interfaces. This PR aims to fix this and enforces the internal use of the openmm unit system. This PR will close issues #54 and #32.
The `forward` call of the models will check for units attached to `positions` and convert them to nanometers if possible. If no units are attached, it will assume that the units are in nanometers and print a warning to stderr.

## Todos
  - [x] Refactor all public APIs to ensure that we check and validate units
  - [x] Conversion to openmm unit system

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -450,9 +450,6 @@ def pair_list(
         i_indices = i_indices[mask]
         j_indices = j_indices[mask]
 
-        # i_indices = i_indices.reshape(-1)
-        # j_indices = j_indices.reshape(-1)
-
     # filter pairs to only keep those belonging to the same molecule
     same_molecule_mask = (
         atomic_subsystem_indices[i_indices] == atomic_subsystem_indices[j_indices]
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -172,19 +172,22 @@ def gaussian_rbf(
     return y
 
 
+from openmm import unit
+
+
 class CosineCutoff(nn.Module):
-    def __init__(self, cutoff: float):
-        r"""
+    def __init__(self, cutoff: unit.Quantity):
+        """
         Behler-style cosine cutoff module.
 
-        Args:
-            cutoff (float): The cutoff distance.
-
-        Attributes:
-            cutoff (torch.Tensor): The cutoff distance as a tensor.
+        Parameters:
+        ----------
+        cutoff: unit.Quantity
+            The cutoff distance.
 
         """
         super().__init__()
+        cutoff = cutoff.value_in_unit_system(unit.md_unit_system)
         self.register_buffer("cutoff", torch.FloatTensor([cutoff]))
 
     def forward(self, input: torch.Tensor):
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -361,7 +361,7 @@ class _GaussianRBF(nn.Module):
         ----------
         n_rbf : int
             Number of radial basis functions.
-        cutoff : float
+        cutoff : unit.Quantity
             The cutoff distance. NOTE: IN ANGSTROM #FIXME
         start: float
             center of first Gaussian function.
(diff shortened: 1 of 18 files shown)


modelforge/tests/test_utils.py
```diff
@@ -89,13 +97,15 @@ def test_gaussian_rbf(RBF):
     assert expected_output.shape == (3, n_rbf)
 
 
-@pytest.mark.parametrize("RBF", [GaussianRBF])
+@pytest.mark.parametrize("RBF", [_GaussianRBF])
 def test_rbf_invariance(RBF):
     # Define a set of coordinates
-    coordinates = torch.tensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]])
+    from openmm import unit
```
chrisiacovella, line 103: Why are you importing openmm units here? 

We had been using openff units in the rest of the code.  Openff-units are directly compatible with openmm units.  `openff.units.openmm` has  `to_openmm` and  `from_openmm` functions if we do need to convert.

modelforge/tests/test_utils.py
```diff
@@ -12,36 +12,39 @@ def test_cosine_cutoff():
...
-    expected_output = torch.tensor([0.5, 0.0, 0.0])
-    output = cosine_cutoff(d_ij, cutoff)
-    assert torch.allclose(output, expected_output, rtol=1e-3)
 
 
 def test_cosine_cutoff_module():
     # Test CosineCutoff module
-    d_ij = torch.tensor([1.0, 2.0, 3.0])
-    cutoff = 2.0
+    from openmm import unit
```
chrisiacovella, line 31: I think we should be using openff-units.

modelforge/tests/test_utils.py
```diff
@@ -73,10 +79,12 @@ def test_gaussian_rbf(RBF):
     assert gaussian_rbf.n_rbf == n_rbf
 
     # Test that the cutoff distance is correct
-    assert gaussian_rbf.cutoff == cutoff
+    assert gaussian_rbf.cutoff == cutoff.value_in_unit_system(unit.md_unit_system)
```
chrisiacovella, line 82: I think we should use openff units, as we've bee using those throughout the code already. 

we would use the syntax:

`assert gaussian_rbf.cutoff == cutoff.is_compatible_with(unit.nanometer)`

chrisiacovella:
We had been using openff-units already in the package.  I don't think we should switch to openmm-units for the public facing API. Openff-units should be more flexible  (as we can easily convert to/from openmm units if we need) and has a convenient function unit.parse_expression, to make it easily to reassign units from the HDF5 files (which are encoded as strings).  

I think we should actually probably switch to openff-units in Chiron, just doing conversions from the openmm test systems (but that is a conversation for a different day in a different repo).

chrisiacovella (changes requested):
as I just commented, I think we should be using openff-units. Maybe we can discuss this quickly offline.  I'm happy to switch these top openff.

chrisiacovella:
After quick discussion, we are going to use openff-units.  I'll be grabbing a copy of the branch and making these changes.

wiederm:
Thanks @chrisiacovella for these changes!

Sorry for introducing yet another unit package! Thanks for catching that in time and making the necessary modifications!

modelforge/potential/utils.py
```diff
@@ -172,28 +172,32 @@ def gaussian_rbf(
...
-        r"""
-        Behler-style cosine cutoff module.
+from openff.units import unit
 
-        Args:
-            cutoff (float): The cutoff distance.
 
-        Attributes:
-            cutoff (torch.Tensor): The cutoff distance as a tensor.
+class _CosineCutoff(nn.Module):
```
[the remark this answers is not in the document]
chrisiacovella: Should this be a public class

chrisiacovella (approved):

PR #68, opened 2024-02-09
Validate schnet family of NNPs 

wiederm:
## Description
This PR validates the implementation of the Schnet family of potentials (SchNET and PaiNN) against its original implementation.

## Todos
  - [x] Compare spk PaiNN with mf PaiNN
  - [x] Compare spk SchNET with mf SchNET 

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -329,7 +329,7 @@ def _shifted_softplus(x: torch.Tensor):
     from torch.nn import functional
     import math
 
-    return functional.softplus(x) - math.log(0.2)
+    return functional.softplus(x) - math.log(2.0)
 
 
 from typing import Optional
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -313,23 +313,30 @@ class EnergyReadout(nn.Module):
         return total_energy_per_molecule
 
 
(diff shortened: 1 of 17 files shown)


modelforge/potential/utils.py
```diff
@@ -329,7 +329,7 @@ def _shifted_softplus(x: torch.Tensor):
     from torch.nn import functional
     import math
 
-    return functional.softplus(x) - math.log(0.2)
+    return functional.softplus(x) - math.log(2.0)
```
chrisiacovella, line 332: Would it be better to keep everything in torch? 

Something like: 

`log_2 = torch.full((x.shape[0],), 2.0, device=x.device).log()
return functional.softplus(x)-log_2
`

Not sure if this would ultimately be more efficient.

chrisiacovella (approved):

modelforge/potential/utils.py
wiederm: I believe that this computation is more difficult because it converts log_2 to a tensor, for which we need to keep track of the dimensions and the device (even though it is only a constant offset).
I have turned this in a nn.Module class, though, and calculate the offset once in the init using math, hopefully removing some of the overhead.

Issue #72, opened 2024-02-28
Difference between atomic self energies from regression and from literature


wiederm:
We need to double-check the source of the relatively significant difference in the atomic self-energies of the QM9 dataset.

chrisiacovella:
I'm (mostly) copying my comment from the PR: 

I wouldn't say these are really all that off.  All but hydrogen have less than a 1% difference.  Hydrogen actually has the smallest numerial difference (delta), but because it is nearly 2 orders of magnitude smaller, it just seems more significant.  

All values in kJ/mol

```
element |  DFT              | regressed         | delta         | % diff
H       |   -1313.466862    |   -1585.320503    |  271.8536413  |  20.69741151
C       |  -99366.70746     |  -99963.18962     |  596.4821665  |   0.600283719
N       | -143309.938       | -143744.2422      |  434.3042769  |   0.303052449
O       | -197082.0672      | -197504.4753      |  422.4081034  |   0.21433107
F       | -261811.5456      | -262199.961       |  388.415491   |   0.148356899
```

But since the regressed values are all consistently higher than the DFT values, if we use them we are going to end up with positive values for small molecules and much smaller energies for large molecules.  That is, the differences are in fact large relative to the scale of the formation energy (which seems to be closer to the scale of the hydrogen atom for a random sampling of molecules from qm9).   But that might just be as accurate as one can get give the scale of the input energies being  typically 10^6 kJ/mol. 

A few quick calculations of the formation energy to demonstrate differences. 

Methane formation energy
DFT self energy: -1656.8639 kJ/mol
regressed self energy: 27.0328 kJ/mol

Ammonia formation energy:
DFT self energy: -1158 kJ/mol
regressed self energy: 91 kJ/mol

water: 
DFT self energy: -891 kJ/mol
regressed self energy: 74.55 kJ/mol

acetylene: 
DFT: -1612 kJ/mol
regressed: 123 kJ/mol

CCC(=O)C=O
DFT: -4922 KJ/mol
regressed: -60 KJ/mol

The larger question: does this actually matter if we can just add the same constant self energy values back on later?

chrisiacovella:
Following up on our discussion today:  we should test this!  

Does it have any impact on the accuracy of the trained model? On the training time? On the stability?  

[...]

Issue #65, opened 2024-02-06
Adding model loaders for SPICE, ANI1x and ANI2x


chrisiacovella:
Curation routines for SPICE, QM9, and ANI1x are already in place; we still need to have a curation routine for ANI2x. 

A dataset (that inherits from HDF5Dataset), has thus far only been set up for QM9; these need to be defined for the other datasets. 

A few notes regarding SPICE:

Currently, we have curation schemes set up for both SPICE 1.1.4 (i.e., the data associated with the paper) and what I've just called "openff" spice. Openff spice is  effectively same data as in 1.1.4, but calculated at the openff level of theory and retrieved from QCArchive.  

Since openff spice is being retrieved from QCArchive, we can easily associated the "source" with each entry (e.g., "SPICE PubChem Set 1 Single Points Dataset v1.2").  This may be very useful for future testing purposes to be able to easily filter out subsets of data. 

The HDF5 file of version 1.1.4 for SPICE includes filtering out of configurations with very high forces. It might be good to also identify those molecules in the openff_spice; we might not want to completely exclude them but rather provide an attribute in the hdf5 file to allow us to remove them if desired

We will also add to openff spice curation a quantity "DFT_total_force"  rather than gradient (so we don't have to change the sign of this at training, and ensure that gradient is not accidentally used in place of force).

PR #74, opened 2024-03-06
Extending and updating curation sets.

chrisiacovella:
## Description
This will add new datasets into model forge, including full spice 1.1.4, preliminary spice 2, ani2x, and the test dataset.   

Notes:  
- Currently, the spice openff code only pulls the pubchem datapoints from qcarchive; the full dataset will be included in the updated code. As part of this, the hdf5 file will get an additional optional source attribute for dataset, to allow for future filtering (as mentioned in #65).  This file will also write out forces, rather than gradients, again as mentioned in #65. 
- As we have been started to discuss in #72 , we do not necessarily need formation energies within the datasets. The linear regression scheme, or using a dictionary of values can be done in the dataset pre-processing.  

## Todos
  - [x] Ani2x
  - [x] ANi2x tests
  - [x] spice 2
  - [x] spice 2 tests
  - [x] full spice 1.1.4
  - [x] Add unit tests for new datasets
  - [x] updated wiki with dataset/hdf5 format description.
  - [x] upon discussion, add field that reports total molecule charge (needed for spice that allow for non-neutral molecules).

## Status
- [x] Ready to go


chrisiacovella:
The wiki has been updated with a lot of examples and discussion about the hdf5 file format and underlying "data" datastructure passed to the hdf5 file. 

https://github.com/choderalab/modelforge/wiki/Dataset-and-curation

chrisiacovella:
I still need to implement the test data set.  I had to rerun some calculations.

chrisiacovella:
There appears to be another change in the naming scheme in one of the datasets (Processing SPICE DES370K Single Points Dataset); I need to add in some regex searching to identify this different naming convention and skip all the sorting by conformers ids.

modelforge/curation/ani2x_curation.py
```diff
@@ -0,0 +1,251 @@
...
+                    # to get the usable mask for the sort, i.e., of shape (n_conformers,),
+                    # we will sum over axis=1 in mask_temp; since a True entry has a value of 1
+                    # we have a match if the sum is equal to the number of atoms in the molecule
+                    mask = np.sum(mask_temp, axis=1) == molecule.shape[0]
+
+                    # to be able to use a dictionary, come up with a string based representation of
+                    # the molecule array, e.g., [ 8, 8 ] becomes "[8_8]"
+                    molecule_as_string = np.array2string(molecule, separator="_")
+
+                    ds_temp["name"] = molecule_as_string
```
wiederm, line 172: I'm not entirely sure why this is necessary. Maybe I'm missing something. Why can't the dictionary value be a NumPy array?

modelforge/curation/ani2x_curation.py
```diff
@@ -0,0 +1,251 @@
...
+                        forces[mask] * self.qm_parameters["forces"]["u_in"]
+                    )
+
+                    self.data.append(ds_temp)
+        if self.convert_units:
+            self._convert_units()
+
+    def process(
+        self,
+        force_download: bool = False,
```
wiederm, line 194: That is a great and necessary option!

modelforge/curation/ani2x_curation.py
```diff
@@ -0,0 +1,251 @@
...
+        unit_testing_max_records: Optional[int] = None,
```
wiederm, line 195: again, that's great for testing!

modelforge/curation/spice_114_curation.py
```diff
@@ -142,6 +147,26 @@ def _init_record_entries_series(self):
...
+        Returns
+        -------
+        total_charge: unit.Quantity
+        """
+
+        from rdkit import Chem
+
+        rdmol = Chem.MolFromSmiles(smiles, sanitize=False)
+        total_charge = sum(atom.GetFormalCharge() for atom in rdmol.GetAtoms())
+        return total_charge * unit.elementary_charge
```
wiederm, line 168: can we cast this to int?

modelforge/tests/test_remote.py
```diff
@@ -25,10 +25,42 @@ def test_is_url():
         == False
     )
 
+
 def test_download_from_figshare(prep_temp_dir):
     url = "https://figshare.com/ndownloader/files/22247589"
     name = download_from_figshare(
         url=url,
+        md5_checksum="c1459c5ddce7bb94800032aa3d04788e",
```
wiederm, line 33: that's great!

modelforge/utils/remote.py
```diff
@@ -61,7 +60,7 @@ def fetch_url_from_doi(doi: str, timeout: Optional[int] = 10) -> str:
     import requests
 
     # force to use ipv4; my ubuntu machine is timing out when it first tries ipv6
-    #requests.packages.urllib3.util.connection.HAS_IPV6 = False
+    # requests.packages.urllib3.util.connection.HAS_IPV6 = False
```
wiederm, line 63: mine too! I had to delete this test previously

modelforge/tests/test_curation.py
```diff
@@ -1450,216 +1481,490 @@ def test_spice114_process_download_conversion(prep_temp_dir):
...
+        local_data_path, filename, unit_testing_max_records=1
+    )
+
+    assert len(ani2x_dataset.data) == 1
+    assert ani2x_dataset.data[0]["name"] == "[1_9]"
+    assert ani2x_dataset.data[0]["n_configs"] == 517
+    assert ani2x_dataset.data[0]["energies"].shape == (517, 1)
+    assert ani2x_dataset.data[0]["atomic_numbers"].shape == (2, 1)
+    assert ani2x_dataset.data[0]["geometry"].shape == (517, 2, 3)
+    assert ani2x_dataset.data[0]["forces"].shape == (517, 2, 3)
```
wiederm, line 1515: Maybe I misremember this, but I thought that the ANI2x dataset contains forces with two levels of theory.

wiederm (approved):
Great work! It's exciting to see how much progress has been made, and I can't wait to start training the models with these new datasets.

modelforge/curation/ani2x_curation.py
chrisiacovella: The dictionary value of course can be a numpy array, but the entry for "name" at some point down the line ends up being used as a dictionary key.  In general, we are assuming that "name" should be a string.  I should update the comment to make sure it is more clear.

modelforge/curation/spice_114_curation.py
chrisiacovella: That would probably be safer.  I've updated it in all the spice curation files

modelforge/utils/remote.py
chrisiacovella: oh, that is an old comment!  I ended up fixing it, it was a weird settings error.  I'm going to leave that import in, commented out, just in case we run into this again.  This is not a very used function anymore, after changing how we structured some of the code.

modelforge/tests/test_curation.py
chrisiacovella: Ani1x has different levels of theory, in the same source hdf5 file.  Ani2x splits them up into separate files.  This is the data from the main paper (and what  is used in the published model).  I need to look at each of the files a bit more closely.  they might be completely interchangeable in terms of the curation code (so we can just add in a flag to set the level of theory).  I think the class for loading in this data would be the same sort of idea, which file it will use, but we need to be better at naming conventions of temporary files.

chrisiacovella: Talked to John about this today;  we are going to just stick with this dataset to start.

Issue #84, opened 2024-03-22
File hashes


chrisiacovella:
When we got to train, we will load up a curated dataset, creating local files which we cache for future use.  We need to store/compare the hash of these files to ensure that we are not working with an incorrect or partially generated file.  This should be trivial for the gzipped hdf5 files, but the npz files we generate locally, we will need to probably generate some metadata file after it is created that stores the hash, since this hash is not known beforehand.

chrisiacovella:
The general sequence of the data loader is 
- download the .hdf5.gz file
- unzip the .hdf5.gz file
- load the .hdf5 file
- save the .hdf5 file as an .npz file
- load the .hdf5 file as an .npz file

Ideally, we want to use the .npz file if it exists, and skip all the rest.   If the .npz doesn't exist, we should check to see if the .hdf5  file exists.  If not, we will check to see if the .hdf5.gz exists.   If not, we will download.  However, we can't just rely on seeing if a file exists, we need to make sure that it is the correct file. 

A few changes I've been making in PR #91 . 

- Rather than using a fully generic filename for the various temporary files (i.e., not using "cached.npz" or something that doesn't tell us which dataset we have) each file has a unique name defined in the data loader (so, something like "qm9_dataset_processed.npz", "qm9_dataset.hdf5", "qm9_dataset.hdf5.gz"). 

- While giving dataset specific names helps, we also need to validate the checksums of the files. The checksums are encoded in the data loader along with the filename.  For example, if we call `_download` we will check to see if the .hdf5.gz exists, and if it does, we will then compare the checksum against the expected checksum.  If the checksum doesn't match, we will know we need to download again.   Note, we can also tell the code to force_download the file.   Similarly, we logged the known checksum for the .hdf5 file. 

[...]

PR #91, opened 2024-03-29
Dataset caching overhaul and additional datasets

chrisiacovella:
## Description
This PR will focus on two key things:
- revamping the dataloader classes to be "Safer" in terms of cached files (checking the MD5 checksum, better naming practices, etc). This refers to issue: #84 
- adding in dataloaders for other datasets (ANI2x, Spice, etc) so the curated datasets can be used in training. 

## Todos
  - [ ] 

## Status
- [ ] Ready to go


chrisiacovella:
At this point still cannot train with this dataset; need to resolve some issues related to default properties.  Currently causes an error due to Q not being defined in ani2x (logic needs to check if "none" is defined...started changing datastructure for properties).

chrisiacovella:
Copying from the issue #84:

The general sequence of loading data is 
- download the .hdf5.gz file
- unzip the .hdf5.gz file
- load the .hdf5 file
- save the .hdf5 file as an .npz file
- load the .hdf5 file as an .npz file

Ideally, we want to use the .npz file if it exists, and skip all the rest.   If the .npz doesn't exist, we should check to see if the .hdf5  file exists.  If not, we will check to see if the .hdf5.gz exists.   If not, we will download.  However, we can't just rely on seeing if a file exists, we need to make sure that it is the correct file. 

A few changes I've been making here:

- Rather than using a fully generic filename for the various temporary files (i.e., not using "cached.npz" or something that doesn't tell us which dataset we have) each file has a unique name defined in the data loader (so, something like "qm9_dataset_processed.npz", "qm9_dataset.hdf5", "qm9_dataset.hdf5.gz"). 

- While giving dataset specific names helps, we also need to validate the checksums of the files. The checksums are encoded in the data loader along with the filename.  For example, if we call `_download` we will check to see if the .hdf5.gz exists, and if it does, we will then compare the checksum against the expected checksum.  If the checksum doesn't match, we will know we need to download again.   Note, we can also tell the code to force_download the file.   Similarly, we logged the known checksum for the .hdf5 file. 

- An issue comes up when dealing with the npz files, as it appears the checksum may be different with different platforms.  Furthermore, the datafile itself might change, if a different set of "properties of interest" are selected.    To handle this, when writing the .npz file, we also write a .json file that contains some metadata, including the checksums of the .hdf5 and .hdf5.gz files used to generate the data, the data_keys used (i.e., properties of interest), as well as the date the .npz file was generated.   We will check to ensure that the checksum of the .hdf5 file used to generated the .npz file matches the checksum in the data loader (to ensure that the code itself hasn't been updated). We also check to ensure that the properties of interest match those defined in the data loader. If all of these match, we can be reasonable safe in our loading.  Note since the .json file is generated AFTER the npz file is written, we will only be able to read this in if the file is properly written.

wiederm:
Thanks for tackling this @chrisiacovella !

How do you resolve the path where the cached files are stored? Are these in userspace or in a scratch/tmp directory?

modelforge/curation/model_dataset.py
wiederm, line 1: That is a very specific use case. I am wondering if we should put this in the curation module or if we should separate this from the package.

modelforge/dataset/model_dataset.py
wiederm, line 1: again, not sure if this is the best place for this.

I think it makes sense to provide a flexible class that can be used for user-defined datasets, but I think in that case, this might be too specific for the package.

scripts/dataset_curation.py
wiederm, line 1: Can you clarify how the curation module interacts with the dataset module? If I want to load the SPICE dataset, is it necessary to interact with the curation module?

modelforge/dataset/dataset.py
```diff
@@ -459,16 +672,48 @@ def _load_or_process_data(
...
+        ) and (
+            data._metadata_validation(
+                data.processed_data_file["name"].replace(".npz", ".json"),
+                data.local_cache_dir,
+            )
+            and not data.force_download
+            and not data.regenerate_cache
+        ):
+
+            data._from_file_cache()
```
wiederm, line 695: very minor comment: `data._from_file_cache() could be moved outside the control flow since each of the branches is calling this method

modelforge/tests/test_dataset.py
wiederm, line 1: Can we add the newly added datasets (only a subset) to the tests here:
https://github.com/chrisiacovella/modelforge/blob/1d4efbe35f06d0cca66a238bffc92799bb4adc10/modelforge/tests/test_dataset.py#L11?

Actually, it would make sense to add these here: https://github.com/chrisiacovella/modelforge/blob/1d4efbe35f06d0cca66a238bffc92799bb4adc10/modelforge/dataset/__init__.py#L5
analogous to 
https://github.com/choderalab/modelforge/blob/0d2f2161580a45c0721171b8a47a57cf72170c98/modelforge/potential/__init__.py#L14

wiederm (approved):
That's a great addition to the package! I left some minor comments and asked for some clarification, but in general I think this is good to go!

chrisiacovella:
> Thanks for tackling this @chrisiacovella !
> 
> How do you resolve the path where the cached files are stored? Are these in userspace or in a scratch/tmp directory?

Each Dataset can be initialized by setting the "local_cache_dir" which will specify where the datafiles end up.

modelforge/curation/model_dataset.py
chrisiacovella: I agree. I'll move these to a public repo on my github.

modelforge/dataset/spice114.py
```diff
@@ -0,0 +1,280 @@
...
+            "Ca": -676.9528465198214 * unit.hartree,  # 2+
+            "Cl": -460.1988762285739 * unit.hartree,
+            "F": -99.78611622985483 * unit.hartree,
+            "H": -0.498760510048753 * unit.hartree,
+            "I": -297.76228914445625 * unit.hartree,
+            "K": -599.8025677513111 * unit.hartree,  # 1+
+            "Li": -7.285254714046546 * unit.hartree,  # 1+
+            "Mg": -199.2688420040449 * unit.hartree,  # 2+
+            "N": -54.62327513368922 * unit.hartree,
+            "Na": -162.11366478783253 * unit.hartree,  # 1+
```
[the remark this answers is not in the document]
chrisiacovella: If we take carbon for example, it could be between about 0.05 and 0.4 hartrees depending on the charged state, or about 30 to 110 kj/mol, roughly speaking.  

`            "C": {-1: -37.91424135791358, 0: -37.87264507233593, 1: -37.45349214963933},
`
That is not necessarily a small difference, but pretty small considering the scale (less than 2% difference in the values). If we did the linear regression on any of the datasets with different subsets of the data, we'd probably have larger variability (as we've talked a bunch about in issue #72).  Taking into this into account would mean we'd have to be passing the charge info when we remove self energies and when we do inference to  add the "right" values back in.  Given the overall goal of removing the self-energies, I don't think this is something we need to worry about.

modelforge/potential/utils.py
```diff
@@ -5,6 +5,8 @@
 import torch
 import torch.nn as nn
 from loguru import logger as log
+from modelforge.utils.units import *
```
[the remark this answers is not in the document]
chrisiacovella: Yeah this is not ideal.  I'm going to switch it up (basically, to be able to convert hartrees to kj/mol I need to add in a different context to openff units, which is just a few lines of code, not a new class.  So I think the easiest is just to set __all__ with a name for this like "chem_context" or something, since I called the context "chem".  That seems to be the simplest way to make sure this is cleaner )

scripts/dataset_curation.py
chrisiacovella: The Curation models just generate the .hdf5 files.  Users will not directly interact with these, they are simply included to preserve how we are transforming the data and generating the hdf5 files for transparency.  This script is just what I used to generate the hdf5 datasets.

modelforge/dataset/dataset.py
chrisiacovella: It could be, but I like having it in each as I think it makes it clearer the exact flow being executed for each conditional block (i.e., keeps the logic self-contained and localized in the file).

modelforge/tests/test_dataset.py
chrisiacovella: I think we need to rewrite/rethink some of these tests.  We run into the issue that the names and even number of  "properties_of_interest" are not consistent among all the the datasets, and thus many of the tests will fail for that reason alone.  

It would be good though to redo this.  I can try to work on that.

chrisiacovella:
@wiederm the equivalent tests seem to be failing for sake on macOS python 3.10.  I resolved these issues earlier by seeding the random number generator in the equivalence_test_utils.  It appears you commented this out (so the tests may not be identical each time).   I think there might be an issue with using Euler angles to generate  the rotation, rather than quaternions.

chrisiacovella:
Ok, tests are all set up now to handle looping over multiple datasets.  I had to cut down a few of the datasets for the test_models; I think we were running out of memory on some of the tests on CI, as they were passing locally (the point is to test the NNP, not the dataset; I still keep qm9 and ani2x so we have some variety).

Issue #96, opened 2024-04-09
Issue generating angular_species12 tensor in ani2x


chrisiacovella:
When I tried to train with the ani2x dataset  with ani2 nnp,  I ran into an issue of indices out of bounds (max bounds should be 6, given there are maximum of 7 atom types).  The translation from original atomic number down to 0 to n_max_atomtypes, seems to be not correct.    This didn't show up trying qm9 with ani2 nnp, because the there are only 5 atom types (max was default to 7).

chrisiacovella:
I think I've tracked this down to  potential.utils.py  (line 135):

```
ATOMIC_NUMBER_TO_INDEX_MAP = {
    1: 0,  # H
    6: 1,  # C
    7: 2,  # N
    8: 3,  # O
    9: 4,  # F
    15: 5,  # P
    16: 6,  # S
    17: 7,  # Cl
    35: 8,  # Br
    53: 9,  # I
}
```

Since the ani nnp has only 7 atoms (H,C,N,O,F, S, Cl), we end up with the numbers being off by one because of "P" in this list.  The reason qm9 worked is that we only  had H, C, N, O, F so we never indexed more than the max number of atom types. 

If I hardcode in this dict into the ani.py model it will train (i.e., just leaving out P, Br, and I):

```
ATOMIC_NUMBER_TO_INDEX_MAP = {
            1: 0,  # H
            6: 1,  # C
            7: 2,  # N
            8: 3,  # O
            9: 4,  # F
            16: 5,  # S
            17: 6,  # Cl
        }
```

So I think we need to think about how to redo this to be flexible.  If we want to use ani with an arbitrary dataset (which many have more atomic species in it), we need to be able to adjust this dict, adjust max number of atomic species, and dynamically set up the atomic networks, i.e.:


```

        self.H_network = atomic_neural_networks["H"]
        self.C_network = atomic_neural_networks["C"]
        self.O_network = atomic_neural_networks["O"]
        self.N_network = atomic_neural_networks["N"]
        self.S_network = atomic_neural_networks["S"]
        self.F_network = atomic_neural_networks["F"]
        self.Cl_network = atomic_neural_networks["Cl"]
```

wiederm:
Thanks for tracking this down @chrisiacovella !

[...]

PR #97, opened 2024-04-09
removed elements not implemented from ATOMIC_NUMBER_TO_INDEX_MAP

wiederm:
## Description
Bugfix for the ANI-2x implementation, addressing issue #96 

## Status
- [x] Ready to go


chrisiacovella (approved):
This is basically what I did (as mentioned in the issue) to get this to work.  This should be a good first fix.  We can discuss later how we want to modify this later to either throw an error if we have elements outside of what is expected or make this more flexible to accommodate any number of elements (which obviously will change the original implementation, but I think that is fine, since are goal isn't to just retrain ani, but use the ani approach with other datasets to compare the result).

PR #99, opened 2024-04-09
Setup Read The Docs

mikemhenry:
Setup RTD :D


mikemhenry:
@wiederm @chrisiacovella I have setup RTD to fail if there are warnings when building the docs, I really recommend this since it will catch things like broken links, BUT it is something you need to remember when you are looking at the build log and think "why does it have an error when there is no error"

chrisiacovella:
I think the CI failures were due to a failure to connect to qcarchive.  I resubmitted one of the tests to see if that was the issue.

chrisiacovella:
This appears to be an issue with qcportal 0.54 (mostly a change in behavior in `ds.entry_names`). I raised an issue with the qcportal folks: https://github.com/MolSSI/QCFractal/issues/817

Basically just need to add in an extra line of `ds.fetch_entry_names()` before querying `ds.entry_names`.

chrisiacovella:
Took more than a single line addition to fix (had to switch to some local caching using the dict of a qcportal record).  But the CI now passing, so I think it is safe to merge.

chrisiacovella (approved):

PR #102, opened 2024-04-16
State Dictionary saving/loading

wiederm:
## Description
We are using a TrainingAdaptor to train neural network potentials (the Pytorch NNP model is an attribute of the TrainingAdaptor, which inherits from Pytorch-Lightning). When we serialize the state_dict, this is reflected in the keys, which are prefixed by `model.`. When we generate a model for inference, we don't use it with the TrainingAdapter, but the vanilla NNP (without any overhead from PyTorch-Lightning). To load a state dictionary correctly, we need to remove the `model.` prefix for each of the state_dict keys.

This PR overrides the native `load_state_dict` and adds a preprocessing step before calling `super().load_state_dict` to remove the `model.` prefix if found in the keys of the state_dict.  

## Todos
  - [x] override native `load_state_dict`
  - [x] write tests demonstrating the correct loading/saving of state_dict

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/__init__.py
+++ b/modelforge/potential/__init__.py
@@ -8,7 +8,6 @@ from .utils import (
     AngularSymmetryFunction,
     FromAtomToMoleculeReduction,
 )
-from .models import TrainingAdapter
 from .models import NeuralNetworkPotentialFactory
 
 _IMPLEMENTED_NNPS = {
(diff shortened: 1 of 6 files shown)


chrisiacovella (approved):
This looks great! I think it can be merged.

PR #106, opened 2024-04-22
Comparing PhysNet against original implementation

wiederm:
## Description
We have yet to compare PhysNet with the original implementation. Since SAKE uses part of the PhysNet implementation now is a good moment to add tests that compare the PhysNet/SAKE/ implementation with the original model.

## Todos
  - [x] test RBF implementation

## Status
- [ ] Ready to go


wiederm:
Reopening this after merging #101

PR #107, opened 2024-04-24
SAKE/PhysNet RBF calculation had difficulties with different units

wiederm:
## Description
The implementation of the RBF calculation of SAKE/PhysNet was written for Angstrom. Since we are operating in nanometer we had to adopt some of the calculations to work properly.

## Todos
  - [x] add test against original PhysNet implementation
  - [x] refactor RBF class
  - [x] update tests  

## Status
- [ ] Ready to go


wiederm:
I believe that the test is failing because of the use of single precision in this test. I have changed this in one of the open PRs. I will merge this for now, since I don't want to change this here again.

wiederm:
Thanks, @ArnNag, for all your work on this! :tada:

PR #108, opened 2024-04-26
Refactor base class for neural network potentials

wiederm:
## Description

Previously, we have implemented all base functionality, including the neighbor list calculation, in a base class from which each neural network inherits. While this was straightforward, this abstraction made exporting the model and exchanging the neighbor list calculation difficult.
To make these use cases work, the new structure is as follows (outlined for ANI):
`ANI2x` class that wraps and initializes `ANI2xCore` and `InputPreparation` instances. The `InputPreparation` class takes cartesian coordinates as input and outputs `d_ij` and `r_ij`, as well as neighbor indices. The `ANI2Core` performs all calculations necessary for the neural network operating on the output of the `InputPreparation`.

When models are exported, only the `Core` class is exported.
 

## Todos
  - [x] Refactor base class
  - [x] Refactor neural network classes
  - [x] Update tests

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/potential/painn.py
+++ b/modelforge/potential/painn.py
@@ -71,7 +71,7 @@ class PaiNNNeuralNetworkData(NeuralNetworkData):
     atomic_embedding: torch.Tensor
 
 
-class PaiNN(BaseNeuralNetworkPotential):
+class PaiNNCore(BaseNeuralNetworkPotential):
     """PaiNN - polarizable interaction neural network
 
     References:
@@ -156,7 +156,6 @@ class PaiNN(BaseNeuralNetworkPotential):
         prior.update(shared_config_prior())
         return prior
 
-
     def _model_specific_input_preparation(
         self, data: "NNPInput", pairlist_output: "PairListOutputs"
     ) -> PaiNNNeuralNetworkData:
@@ -207,9 +206,9 @@ class PaiNN(BaseNeuralNetworkPotential):
         mu = transformed_input["mu"]
         dir_ij = transformed_input["dir_ij"]
 
-        for i in [1,2,3]:
+        for i in [1, 2, 3]:
             print(i)
-        
+
         for i, (interaction_mod, mixing_mod) in enumerate(
             zip(self.interaction_modules, self.mixing_modules)
         ):
@@ -232,7 +231,6 @@ class PaiNN(BaseNeuralNetworkPotential):
             "q": q,
             "atomic_subsystem_indices": data.atomic_subsystem_indices,
         }
-
 
 
 from openff.units import unit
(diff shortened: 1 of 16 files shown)


modelforge/potential/painn.py
```diff
@@ -207,9 +190,9 @@ def _forward(
         mu = transformed_input["mu"]
         dir_ij = transformed_input["dir_ij"]
 
-        for i in [1,2,3]:
+        for i in [1, 2, 3]:
             print(i)
```
chrisiacovella, line 194: Remove the print statement.

modelforge/potential/processing.py
```diff
@@ -0,0 +1,339 @@
+import torch
+from modelforge.dataset.dataset import DatasetStatistics
+from loguru import logger as log
+
+
+class FromAtomToMoleculeReduction(torch.nn.Module):
```
chrisiacovella, line 6: While this function is generic in what it does (summing the per atom quantities) all the variable names and description refer to energy (obviously this is what we are using it for in the code).  Not sure if it makes sense to rename the class to like "FromAtomToMoleculeEnergyReduction".  Obviously this is minor, but given the ever increasing complication of the code, being more specific could be better.

modelforge/tests/test_training.py
```diff
@@ -1,8 +1,10 @@
-from typing import Type
-
+import os
 import pytest
 
+IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
+
 
+@pytest.mark.skipif(IN_GITHUB_ACTIONS, reason="Skipping this test on GitHub Actions")
```
chrisiacovella, line 7: I'm assuming this just takes too long on CI?

chrisiacovella:
Can you write out a slightly more detailed description of the hierarchy?  Might be good to just copy and paste the same text to an issue and the wiki.

chrisiacovella (approved):
Just a few, non-blocking comments.  As I said, I think a slightly longer description of the hierarchy would be good.

wiederm:
I outlined the structure here:
https://github.com/choderalab/modelforge/wiki/Neural-network-potentials#module-organization

It might be enough to add the URL?

modelforge/potential/processing.py
wiederm: I think it might be better to move in the opposite direction and remove all references to `energy`. It should be generic, then we don't need to add another class to perform the same calculation for some other property.

modelforge/tests/test_training.py
wiederm: it ran out of memory on macos, it wouldn't take too long. I will add a more specific 'skip if on MacOS statmente'

PR #114, opened 2024-05-09
Check that for hydrogen atoms that are chemically and physically equivalent the atomic energies are the same

wiederm:
## Description

Updated description reflecting @jchodera comment:
This test checks for equivalent atomic representations (and, therefore, atomic energies) of atoms in symmetric geometries that are chemically and physically equivalent. This is tested on the hydrogen atoms of methane with Td symmetry and the hydrogen atoms of ammonia with trigonal pyramidal geometry (C3v).

## Status
- [x] Ready to go


jchodera:
This is only true for geometries that possess symmetry, is it not? For methane, the geometry needs to satisfy the tetrahedral point group (Td). For ammonia, it would have to be C3v. Arbitrary geometries will not satisfy these requirements and can have atomic contributions to the energy that differ.

If you aren't able to guarantee geometry, perhaps something like H2 is a better test?

wiederm:
Yes, that is correct. I will update the comment to reflect that --- the test uses the geometries that satisfy the symmetry groups.

wiederm:
To finish this PR we need to address issue https://github.com/choderalab/modelforge/issues/115 first

wiederm:
Closing here, will reopen with current main

PR #118, opened 2024-05-11
Addressing overuse of fixtures

wiederm:
## Description

We have accumulated technical debts in the configuration of some of our tests. This PR is a first attempt to clean up the overuse of slightly similar but different fixtures, for which the control flow is hidden in `conftest.py,` making it sometimes difficult to understand what is done in a particular test.

I opted for explicitly passing control parameters to each test instead of hiding these in fixtures. I have cleaned up some of the fixtures and simplified them where appropriate. We still use fixtures to set up datasets/databases/database containers. (Sidenote: In one of the next PRs, we need to address the inflationary use of the word dataset in the classes --- it is challenging to distinguish between some of the concepts without reading the documentation.)

## Todos
  - [x] simplify fixtures
  - [x] clean up tests

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -62,22 +62,24 @@ class TorchDataset(torch.utils.data.Dataset[Dict[str, torch.Tensor]]):
         self.dataset_statistics: Dict[str, float] = {}
 
         self.properties_of_interest["atomic_numbers"] = torch.from_numpy(
-            dataset[property_name.Z]
-        )
+            dataset[property_name.Z].flatten()
+        ).to(torch.int32)
         self.properties_of_interest["positions"] = torch.from_numpy(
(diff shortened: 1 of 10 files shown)


wiederm:
@chrisiacovella , can you please look at the two tests still failing in the dataset test module? I can't figure out why these are not passing, I think it has to do with the caching.

wiederm:
I think that one is ready for review! @chrisiacovella , could you please take a look at the failing tests for the dataset, I can't seem to figure out what is going wrong here.

wiederm:
Merging `main` into this branch led to some test failures that went undetected with the current test coverage in `main`. This PR extends the test coverage to all datasets and potentials that are implemented for most tests (there are exceptions) and clearly indicates which of them are used in a given test.

Issue #120, opened 2024-05-16
Removing self-energies is costly


chrisiacovella:
As the title suggests, the process of removing the self-energies from a dataset can be prohibitively expensive with the current implementation as the size of the dataset grows. E.g., QM9 is small and only takes about 90 seconds, whereas ANI2X takes 2 and half hours. 

Since most of the datasets are composed of many conformers that will have identical self energies, the simplest way to speed this up will be to just add a dictionary and avoid recalculating if we have already seen a molecule with the same atomic numbers.  

Implementing this takes ANI2X down about 20 minutes which I think is perfectly manageable.  I think it would still be good to avoid having to redo this calculation at all, by writing this out (instead of writing to local_cache_dir, I think it would be better to be "local_working_dir"...where local_cache_dir is for the raw datasets that we may be sharing among calculations , and local_working_dir is where we pieces specific to the calculation itself) . There already is a _to_cache function in DataModule.prepare_data that writes out the .pt file with the self energies removed, however, this has fixed file name for all datasets ( so we could run into some issues) and  being saved to the execution directly, not to any sort of specified directory (working or cache).  

If we skip calling `prepare_data` and just call `setup`, it looks like it will use the previously written .pt file, but I think we might need to lightly revamp the flow those functions to possible automatically check if the .pt file exists (with options to overwrite, also with appropriate metadata like we do for the .npz file).  Does it make sense to add another layer in the  data loading step to check if there is a .pt file (and check metadata), before looking for the .npz file (then, looking for .hdf5, then looking for .hdf5.gz).  

I'm going to push a PR with the simple change to the self energy removal to using a dictionary to speed things up; we ca discuss and address the other aspects separately.

wiederm:
Thanks for raising this! For a dataset like SPICE, we should aim for a < 5 min preprocessing cost if possible (assuming a multicore CPU with 4 cores). I think caching is a great idea --- do you think we can use the `cache` decorator/method from [functools](https://docs.python.org/3/library/functools.html#functools.cache)?

wiederm:
To bring this up here in the issue, this is the function that we want to speed up:

```python
for i in range(len(dataset)):
    atomic_numbers = list(dataset[i]["atomic_numbers"])
    E = dataset[i]["E"]
    for Z in atomic_numbers:
        E -= self_energies[int(Z)]
        dataset[i] = {"E": E}
```

wiederm:
[...]

PR #121, opened 2024-05-16
Improved the efficiency of the `_subtract_self_energies` function.

chrisiacovella:
This PR provides improved efficiency of the `_subtract_self_energies` function in the `DataModule` (in dataset.py).  As mentioned in issue #120, this step was prohibitively expensive for large datasets (2.5 hours on my laptop for ANI2X, over 3 hours on the lilac cluster).  This is a simple fix that stores molecule self energy in a dictionary,  so we do not need to recompute for each conformer of a molecule.   The dictionary key for this is the concatenation of the atomic numbers into a string (separated by spaces); this does not differentiate between different molecules with the same atoms in the same order (but different structures), which is fine, because our calculation of the self energy is just the sum of atomic self energies and does not consider any other possible aspects of the molecule. 

Also,  I switched to summing a numpy array of the energies, rather than summing via a for loop,  also seems to provide roughly a 10% speed up as well. 

This reduces the time for remove self energies for ANI2X  from 2.5 hours to ~20 minutes on my local machine.   For SPICE2 this takes about 11 minutes, so I'm not sure if it's worth doing any further optimization beyond this (e.g., multithreading); Improving the caching to avoid rerunning step altogether is likely a better focus. 

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -969,44 +969,11 @@ class DataModule(pl.LightningDataModule):
         from tqdm import tqdm
 
         log.info("Removing self energies from the dataset")
-
-        # For efficiency, this will batch conformers with the same number of atoms,
-        # creating a 2d tensor of atomic numbers with shape (N_conf, N_atoms).
-        # The cost of performing this calculation on a tensor of shape (N_conf,N_atoms) is
-        # less the total cost of performing a calculation on a tensor of shape (N_atoms,) N_conf times.
-        # For the tested datasets, this works out to about a 25% savings.
-
-        last_shape = dataset[0]["atomic_numbers"].shape[0]
-        batch = []
         for i in tqdm(range(len(dataset)), desc="Removing Self Energies"):
-
-            if dataset[i]["atomic_numbers"].shape[0] == last_shape:
-                batch.append(i)
-            else:
-
-                atomic_numbers_batch = torch.stack(
-                    [dataset[j]["atomic_numbers"] for j in batch]
-                )
-
-                energies = torch.sum(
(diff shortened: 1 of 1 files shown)


modelforge/dataset/dataset.py
```diff
@@ -953,13 +964,21 @@ def _subtract_self_energies(self, dataset, self_energies: Dict[int, float]) -> N
...
 
+        stashed_values = {}
         log.info("Removing self energies from the dataset")
         for i in tqdm(range(len(dataset)), desc="Removing Self Energies"):
+
             atomic_numbers = list(dataset[i]["atomic_numbers"])
+            name = " ".join([str(int(x)) for x in atomic_numbers])
+
+            if name not in stashed_values.keys():
+                energy = np.array([self_energies[int(Z)] for Z in atomic_numbers]).sum()
```
wiederm, line 975: great addition!

wiederm (approved):
LGTM!

chrisiacovella:
I'm working through some ways still to speed this up.  Switching to using the tensor approach  makes the performance the same as using a simple dictionary to cache (speeds up each calculation enough that it is the same as the overhead associated with reducing the number of calculations).  Using caching with either a dictionary or using functools along with the faster tensor sum makes almost no difference; the overhead of caching seems to negate any performance gains at this point. 

  If I do batching based on the size of the atomic indices (still using the tensor sum, just stacking together all the same sized molecules), this seems to save another 25%. Simply looping over the dataset seems to be costly, so despite not really wanting to go down the multithreading route, I think that is ultimately what we need to do if we want to speed this up any faster.

PR #122, opened 2024-05-17
Speed up preprocessing for dataset

wiederm:
## Description
There are two possible bottlenecks for the preprocessing pipeline:
- Updating the mean and variance estimators
- Removing self energies

In the following, I will outline some of the operations that I noticed are responsible for these bottlenecks. Not all of them can be easily mitigated, but there is room for further improvement.

## Todos
  - [ ] TODO 1

## Status
- [ ] Ready to go


wiederm:
Closing this since @chrisiacovella already prepared a PR with the same scope in #123

PR #123, opened 2024-05-17
Further speed up to _subtract_self_energies 

chrisiacovella:
Further speed up to _subtract_self_energies in DataModule by directly indexing into the atomic_numbers tensor to avoid additional overhead.  Reduces

As further described in Issue #120 , the overhead of `__getitem__` seems to be an additional factor.  We only are using the atomic numbers, so all the other data returned when indexing into the dataset are not needing and just slowing things down.  I just directly indexed into the tensor to get the relevant atomic numbers, and that seems to give some further substantial speed ups (reducing cost about ~50%).  Update: Overlooked that I was using the getter to retrieve energy.  Speed up is even better (see timings below). 

This will additionally address issue #124, swapping in np.cumsum for torch._utils._accumulate in the record splitting routine. 

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -978,8 +978,7 @@ class DataModule(pl.LightningDataModule):
                     dataset.properties_of_interest["atomic_numbers"][start_idx:end_idx]
                 ]
             )
-            dataset[i] = {"E": dataset[i]["E"] - energy}
-
+            dataset[i] = {"E": dataset.properties_of_interest["E"][i] - energy}
         return dataset
 
     def train_dataloader(self) -> DataLoader:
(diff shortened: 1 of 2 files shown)


chrisiacovella:
I realized I was still using the getter function for fetching the energy.  Switching this to direct access further reduced the time.  SPICE2 is now 54 seconds, ANI2X, 4:36.

chrisiacovella:
@wiederm Just noticed that you had started to make some of the same changes in #122.

wiederm (approved):
That looks really good and gives great speed improvement.

I think we will also need to adopt this for the training/validation loop, since the `__getter__` is called repeatedly and, while it might not be the most important part, it might add a significant overheard to each batch processing.

PR #126, opened 2024-05-19
Adapt training for forces

wiederm:
## Description
Tests still need to cover the Loss functions used in the training routine, this PR addresses this.

This PR also adds forces to the getter method of the `TorchDataset` and the `collateconformers` method, as raised here: #127 . It also improves performance by casting dtype on the full tensor and not on individual tensor slices.

## Todos
  - [x] Add test for loss energy/force

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -206,9 +206,9 @@ class TorchDataset(torch.utils.data.Dataset[Dict[str, torch.Tensor]]):
         positions = self.properties_of_interest["positions"][
             series_atom_start_idx:series_atom_end_idx
         ]
-        E = self.properties_of_interest["E"][
-            idx
-        ]  # NOTE: upgrading to float64 to avoid precision issues
+        E = self.properties_of_interest["E"][idx]
+        F = self.properties_of_interest["F"][series_atom_start_idx:series_atom_end_idx]
+
         Q = self.properties_of_interest["Q"][idx]
 
         return {
@@ -216,6 +216,7 @@ class TorchDataset(torch.utils.data.Dataset[Dict[str, torch.Tensor]]):
             "positions": positions,
             "total_charge": Q,
             "E": E,
+            "F": F,
             "atomic_subsystem_counts": torch.tensor([atomic_numbers.shape[0]]),
             "idx": idx,
         }
@@ -1030,6 +1031,7 @@ def collate_conformers(conf_list: List[Dict[str, torch.Tensor]]) -> "BatchData":
 
     Z_list = []  # nuclear charges/atomic numbers
     R_list = []  # positions
+    F_list = []  # forces
     E_list = []  # total energy
     Q_list = []  # total charge
     atomic_subsystem_counts = []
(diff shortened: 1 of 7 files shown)


modelforge/tests/test_training.py
```diff
@@ -57,6 +53,34 @@ def test_train_with_lightning(model_name, dataset_name):
...
+    model = NeuralNetworkPotentialFactory.create_nnp("inference", model_name)
+
+    from modelforge.train.training import EnergyAndForceLoss
+    import torch
+
+    loss = EnergyAndForceLoss(model)
+
+    r = loss.compute_loss(next(iter(dm.train_dataloader())))
+
+    loss = EnergyAndForceLoss(model, include_force=True)
```
wiederm, line 79: @chrisiacovella , can you please take a look at the test failure for the datasets in this test? It seems as if there are indexing errors for the datasets (both for the energy and the force loss)

wiederm: The errors for the energy loss for the different variants of the SPICE are caused by a mismatch between the elements covered by the ANI network and the elements present in the dataset. I think this is fine, but we should through a more explicit and understandable error than a `IndexError` (which obscures the underlying cause).

But I can't figure out the indexing error in the force loss.

chrisiacovella: @wiederm I'm going to additional set up a dataset for spice that only includes the elements supported by ANI (both a "full" and "testing" dataset).  I'm trying to decide the best way to handle this. I think it might be easy to put a filtering option in the dataset curation, like I've done for addressing limiting the number of conformers for the training set generation.  I initially thought having this be a post processing step, which would be fine for the full datasets, but not for test datasets that are meant to have a specific number of conformers.

modelforge/dataset/dataset.py
```diff
@@ -206,16 +206,17 @@ def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
         positions = self.properties_of_interest["positions"][
             series_atom_start_idx:series_atom_end_idx
         ]
-        E = self.properties_of_interest["E"][
-            idx
-        ]  # NOTE: upgrading to float64 to avoid precision issues
+        E = self.properties_of_interest["E"][idx]
+        F = self.properties_of_interest["F"][series_atom_start_idx:series_atom_end_idx]
```
chrisiacovella, line 210: This is assuming that all datasets have forces.  This will fail for qm9.    I had made F and Q optional in the PropertyNames class...we could always change the data loaders to just pack the force with zeroes if not available, or we I guess need to make this some how optional here (maybe this will look over the PropertyNames to know what is available first?)

chrisiacovella: Maybe this ends up just going back to @jchodera point of dumping QM9 altogether!  But, I think we still want some ability to be able to comparing training with and without forces for comparison.  So I'm not sure how best to handle this. Maybe we can discuss Tuesday morning.

chrisiacovella (changes requested):
I think we just need to discuss how to handle datasets without forces, while still being able to exclude forces.

modelforge/dataset/dataset.py
wiederm: Actually, the tests are passing just fine with QM9. In that case all forces are zero, but I think that is fine.

wiederm: There will be an error thrown when trying to calculate the loss, since there are no QM forces, though:
https://github.com/choderalab/modelforge/blob/f98cd99f192ecbea32c428cc975e6f506478ff2e/modelforge/train/training.py#L49

One possible strategy is also to initialize these with zeros. Then no error is thrown but the force contributions are ignored in the loss calculation.

modelforge/tests/test_training.py
wiederm: that is a great idea!

The problem is only for ANI, though; none of the other potentials are limited to a set of elements. So I think it's also fine to just don't allow this combination of model and dataset.

PR #128, opened 2024-05-22
Refactor training pipeline

wiederm:
## Description
Reproducing the published performance metrics with the implemented neural networks is difficult. Even for networks that match their original implementation, there is variability in the training routine that seems to effect the training performance significantly. e.g., the reported training performance of SchNet on QM9 is ~0.3 kcal/mol, while the modelforge SchNet currently has an [average test set performance](https://github.com/choderalab/modelforge/wiki/Experiments) of 1.5 kcal/mol.

In this PR I am working through the different approaches that have been reported to improve training performance.

### Normalize dataset

Talking with colleagues and reading this [work](https://arxiv.org/abs/2308.08859) one issue that is often raised is the energy unit dependency of training performance. The hypothesis is that the closer the distribution of $y$ is matched by the initial distribution of the network parameters the easier it is to converge training. Since we opted for kJ/mol, and the actual units don't matter, I will normalize energies for training (dividing by the standard deviation of the energy across the dataset and subtracting the energy mean).

Considerations:
- We opted to directly modify the dataset **before** training as the general approach. Initially, the only operations that were performed were shifting operations (e.g., subtracting atomic self-energies). This had no further impact on calculations besides a gain in efficiency. By normalizing energies units are changed. We can calculate $\hat{E}$  and $L(\hat{E}, E)$ on the normalized energies, but as soon as we mix these with a force loss it is unclear how to best handle this. I think the best approach to avoid any further complications is to calculate $\Delta(\hat{E},E)$ on the normalized energies, but then multiply by the dataset stddev. Also, if we calculate the force loss we perform this calculation on (\hat{E} * E_stddev).

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -1008,6 +1008,7 @@ class DataModule(pl.LightningDataModule):
             batch_size=self.batch_size,
             collate_fn=collate_conformers,
             num_workers=4,
+            shuffle=True,
         )
 
     def val_dataloader(self) -> DataLoader:
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -855,7 +855,7 @@ class DataModule(pl.LightningDataModule):
 
         # normalize dataset
         if self.normalize:
-            self._normalize_dataset(torch_dataset)
+            self._normalize_dataset(torch_dataset, stats)
 
         # Save processed dataset and statistics for later use in setup
         self._cache_dataset(torch_dataset)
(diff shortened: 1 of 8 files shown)


wiederm:
It appears as if the difference in training performance reported in the description of this PR can be attributed to the default behavior of the PyTorch dataset, which has been changed in PR #129 .

I will port some of the changes proposed in this PR in a new, separate PR, but I would like to avoid normalization by default if possible. I will leave this PR open in the meantime.

wiederm:
After multiple training repeats, I can confirm that `modelforge` SchNet approaches the reported MAE in the test set performance without normalizing energies.

PR #137, opened 2024-05-31
Training options and control parameters moved to toml files

wiederm:
## Description
This PR refactors the training script and moves the training functions to the module. It also extracts parameters set in the script and moves them to a toml file. The structure of the toml file is as follows, each with a `potential`, `training` and `dataset` entry.

```toml
[potential]
potential_parameters.max_Z = 101
potential_parameters.number_of_atom_features = 32
potential_parameters.number_of_radial_basis_functions = 20
potential_parameters.cutoff = "5.0 angstrom"
potential_parameters.number_of_interaction_modules = 3
potential_parameters.shared_interactions = false
potential_parameters.shared_filters = false
model_name = "SchNet"

[training]
nr_of_epochs = 50
save_dir = "test"
experiment_name = "your_experiment_name"
accelerator = "cpu"
training_parameters.include_forces = false
num_nodes = 1
devices = 1                                #[0,1,2,3]
early_stopping.monitor = "rmse_val_loss"
early_stopping.min_delta = 0.001
early_stopping.patience = 50
early_stopping.verbose = true

[dataset]
dataset_name = "QM9"
batch_size = 512
remove_self_energies = true
```

Currently, we pass the config dictionaries derived from this toml file to the `create_nnp` method of the `PotentialFactory`. This will internally use dictionary unpacking to pass by argument (without kwargs) to the dedicated functions. All default values of these functions have been removed. In that way we enforce that *all* possible keys are defined in the toml file and ensure that keys are spelled correctly (and no unintended default values are used).

## Status
- [x] Ready to go


wiederm:
@chrisiacovella , what do you think about these changes? we have discussed this in parts already, but I think this should make it easier to track the experiments with dvc and pytorch-lightning.

I am a bit concerned that we are hiding function calls and using toml/yaml files for defining parameters passed to these functions. The drawback of this approach is that there is no obvious way to understand the parameters and the user (and we) need to inspect the code to access the docstring (because the function call is hidden). Do you have a suggest to address this?

PR #138, opened 2024-06-01
Refactor TorchDataset.__getitem__ to return BatchData instead of a dictionary for individual slices

ArnNag:
## Description
I think it makes more sense for `TorchDataset.__getitem__` to return the same type as the DataLoader rather than a custom defined dictionary, as this avoids the need to define redundant abstractions. Thus, `collate_conformer` now takes in a list of `BatchData` instances and collates them into a single `BatchData` instance. This required me to move `BatchData`, `NNPInput`, and `Metadata`  to `modelforge.dataset.dataset` to avoid circular imports; I also feel that placing these classes there makes more sense than placing them in `modelforge.potential.utils`.

# Questions
- [ ] @wiederm I don't understand the Ray tuner reporting and why this is breaking it.

## Status
- [ ] Ready to go


wiederm:
The changes introduced by PR #146 have resulted in merge conflicts. I will review on Monday and resolve conflicts.

wiederm:
I moved the PR to #154

PR #142, opened 2024-06-03
Profiling scripts for training/inference

wiederm:
## Description
Training on larger dataset has high cost associated with each epoch (e.g., ANI2x trained on the ANI2x dataset currently takes ~8h for a single epoch on a GTX 3090). Before committing, we should spend a few days profiling the training pipeline.

## Todos
  - [x] Profiling script for model forward/backward pass (using PyTorch Profiler)
  - [x] Profiling a training scenario (using PyTorch Lightning Profiler)
  
## Status
- [ ] Ready to go

### DIFF
--- a/scripts/profile_epoch.py
+++ b/scripts/profile_epoch.py
@@ -54,7 +54,7 @@ def setup(model_name: str):
         num_nodes=num_nodes,
         devices=devices,
         accelerator=accelerator,
-        profiler="simple",  # "advanced",
+        profiler="pytorch",  # "advanced",
     )
 
     dm.prepare_data()
--- a/scripts/profile_epoch.py
+++ b/scripts/profile_epoch.py
@@ -1,0 +1,81 @@
+import torch
+from modelforge.train.training import return_toml_config
+from typing import Dict, Any
+
+
+def perform_training(trainer, model, dm):
+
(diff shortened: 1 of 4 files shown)


chrisiacovella (approved):
I think this is a good start.  I'm doing a bit more digging into the profiler (really how to interpret the output in terms of multicore systems.  Do not want to misinterpret the results and waste our time on the wrong parts).

scripts/profile_epoch.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def setup(model_name: str):
+    from modelforge.dataset.utils import RandomRecordSplittingStrategy
+    from lightning import Trainer
+    from modelforge.potential import NeuralNetworkPotentialFactory
+    from modelforge.dataset.dataset import DataModule
+
+    config = return_toml_config(
+        f"../modelforge/tests/data/training_defaults/{model_name.lower()}_qm9.toml"
```
chrisiacovella, line 23: I think we should change this to use module specified paths, not relative. 

So something like

```
from importlib import resources
from modelforge import tests as modelforge_tests


config = return_toml_config(
        f"{resources.files(modelforge_tests)}/data/training_defaults/{model_name.lower()}_qm9.toml"
)
```

PR #145, opened 2024-06-03
Fixed error in spice2.yaml file.  

chrisiacovella:
Added in test to check dataset download.  The tests will be skipped ( only will run if --run_data_download is passed to the CLI).  This test will take a lot of time so I'm sure this is not something we would want to do with any frequency. It might be worth creating separate tests (instead of parameterize) to make it easier to just specifically test a given dataset. 

It might also be better to make this NOT a pytest test and just a separate piece of code that gives more options to check/test very specific datasets and versions and give more descriptive outputs.  Having something like that in the dataset module, where people can more easily query available datasets and versions. 

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/dataset/yaml_files/spice2.yaml
+++ b/modelforge/dataset/yaml_files/spice2.yaml
@@ -4,29 +4,29 @@ latest_test: nc_1000_v0
 full_dataset_v0:
   version: 0
   gz_data_file:
-    length: 26313472231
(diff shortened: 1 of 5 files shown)


wiederm:
That PR has high priority; the CI tests for all other PRs are failing since some of the hashes for the datasets have changed in the remote location. @chrisiacovella , is this PR ready for review?

Issue #141, opened 2024-06-03
Move part of neighborlist calculation to the dataloader/dataset 


wiederm:
The neighbor list calculations (calculating the local neighborhood N(i) for each i, the distance vector r_ij between atom pairs, and the distance d_ij) are repeated in each epoch. We should move this calculation to the dataloader and cache the results.

wiederm:
A priori it isn't easy to know how much overhead the neighbor list calculation adds. To understand this better I passed a batch of data through the model and also performed the backward pass with the PyTorch Profiler. Below are the stack trace sorted (1) by total CPU time and (2) total GPU time. The neighborlist calculation are by far the most expensive calculations.

Sorted by total CPU time:


```
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ---------------------------------------------------------------------------  
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg     Self CUDA   Self CUDA %    CUDA total  CUDA time avg    # of Calls  Source Location                                                              
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ---------------------------------------------------------------------------  
                                            aten::index        23.65%      54.404ms        41.11%      94.565ms      30.040us      41.266ms        16.28%     101.261ms      32.167us          3148  modelforge/potential/models.py(64): enumerate_all_pairs                      
                                                                                                                                                                                                     modelforge/potential/models.py(216): forward                                 
                                                                                                                                                                                                     torch/nn/modules/module.py(1534): _call_impl                                 
                                                                                                                                                                                                     nn.Module: Neighborlist_0                                                    
                                                                                                                                                                                                     modelforge/potential/models.py(516): prepare_inputs                          
                                                                                                                                                                                                                                                                                  
                                           aten::arange         9.96%      22.908ms        13.90%      31.988ms      10.161us      18.577ms         7.33%      38.867ms      12.347us          3148  <built-in method arange of type object at 0x7f887bb60c20>                    
                                                                                                                                                                                                     modelforge/potential/models.py(64): enumerate_all_pairs                      
                                                                                                                                                                                                     modelforge/potential/models.py(216): forward                                 
                                                                                                                                                                                                     torch/nn/modules/module.py(1534): _call_impl                                 
                                                                                                                                                                                                     nn.Module: Neighborlist_0                                                    
[...]

PR #146, opened 2024-06-06
Splitting pairlist and neighborlist calculation

wiederm:
## Description
As discussed in issue #141 training takes too long and spends most of the compute time generating the pairlist. To move this from a per-epoch cost to a one-time cost this PR moves the calculation of the pairlist to the dataset and saves the pairlist. 
The pairlist is then passed, as an optional parameter, as input to the network. The Neighborlist class takes it as optional input and calculates distances to further filter it to the local neighborhood for each atom within a cutoff. 

## Todos.
  - [x] Cache pairlist calculation
  - [x] Add tests

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/tests/test_models.py
+++ b/modelforge/tests/test_models.py
@@ -398,7 +398,7 @@ def test_pairlist():
     from modelforge.potential.models import Pairlist, Neighborlist
     import torch
 
-    atomic_subsystem_indices = torch.tensor([80, 80, 80, 11, 11, 11])
+    atomic_subsystem_indices = torch.tensor([0, 0, 0, 1, 1, 1])
     positions = torch.tensor(
         [
             [0.0, 0.0, 0.0],
--- a/modelforge/tests/test_models.py
+++ b/modelforge/tests/test_models.py
@@ -258,8 +258,12 @@ def test_forward_pass(
     if "JAX" not in str(type(model)):
 
         # assert that the following tensor has equal values for dim=0 index 1 to 4 and 6 to 8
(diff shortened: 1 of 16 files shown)


modelforge/tests/test_models.py
```diff
@@ -37,20 +47,39 @@ def test_model_factory(model_name, simulation_environment):
     )
     from modelforge.train.training import TrainingAdapter
 
+    # read default parameters
+    from modelforge.train.training import return_toml_config
+
+    config = return_toml_config(
+        f"modelforge/tests/data/potential_defaults/{model_name.lower()}_defaults.toml"
```
chrisiacovella, line 54: This will only work if you run pytest from modelforge's root directory.  We should update this to get the full path using resources (I will push a change for this) and also make sure that .toml files are included in the setup file for pip so they are copied over to the egg.

modelforge/tests/test_models.py
```diff
@@ -37,20 +47,39 @@ def test_model_factory(model_name, simulation_environment):
...
+        model_type=model_name,
         simulation_environment=simulation_environment,
+        model_parameters=potential_parameters,
     )
     assert (
         model_name.upper() in str(type(model)).upper()
         or "JAX" in str(type(model)).upper()
     )
-
+    config = return_toml_config(
```
chrisiacovella, line 70: same comment here about getting the full path.

modelforge/tests/test_models.py
wiederm: thanks! that is a good point!

chrisiacovella (approved):

PR #150, opened 2024-06-08
Refactor pairlist representation

wiederm:
## Description
Instead of padding the pair list to generate an (M,N,2) tensor with M number of molecules, N maximum number of pairs in the dataset, and 2 the pairlist dimension we can generate a (K,2) tensor with K as the total number of atom pairs in the dataset. This requires keeping track of index arrays to correctly retrieve atom pairs for a molecule.
 
## Todos
  - [x] Update atom pair representation, index arrays and offsets
  - [x] Add tests

## Status
- [x] Ready to go


wiederm:
With these changes, we are able to decrease training time to ~15 min / epoch on a single GTX RTX 3090 for the ANI2x dataset (training ANI2x).

chrisiacovella (approved):

PR #152, opened 2024-06-10
Ref ani neighborlist

wiederm:
## Description

This PR makes sure that the ANI model reduces a pairlist/neighborlist with redundant atom pairs (i.e., atom indices for the pair atom 1 and atom 2 are ((1,2),(2,1))). For ANI, we require non-redudant atom pairs. 
There are multiple points where the different pairlist could be implemented. 
I opted for:
(1) a keyword provided to the dataset generation that controls whether to include redundant pairs or not
(2) inside the ANI model an additional indexing step that will reduce a redundant to a non redundant pairlist

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/tests/test_dataset.py
+++ b/modelforge/tests/test_dataset.py
@@ -455,7 +455,6 @@ def test_dataset_neighborlist(model_name, single_batch_with_batchsize_64):
     """Test the neighborlist."""
 
     nnp_input = single_batch_with_batchsize_64.nnp_input
-    nr_of_mols = nnp_input.atomic_subsystem_indices.unique().shape[0]
 
     # test that the neighborlist is correctly generated
     # cast input and model to torch.float64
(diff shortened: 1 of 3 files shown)


modelforge/tests/test_dataset.py
```diff
@@ -484,89 +483,84 @@ def test_dataset_neighborlist(model_name, single_batch_with_batchsize_64):
...
-        # test the pairlist for the ANI potential (whith non-redundant atom pairs)
-        # TODO
-        pass
-    else:
-        # test the pairlist for message passing networks (with redundant atom pairs)
-        # first molecule is methane, check if bonds are correct
-        methan_bonds = pair_list[:, :20]
+    # test the pairlist for message passing networks (with redundant atom pairs)
+    # first molecule is methane, check if bonds are correct
+    methan_bonds = pair_list[:, :20]
```
chrisiacovella, line 488: just a minor typo, might want this to be `methane_bonds`

modelforge/tests/test_dataset.py
```diff
@@ -484,89 +483,84 @@ def test_dataset_neighborlist(model_name, single_batch_with_batchsize_64):
...
 
-        assert (
-            torch.any(
-                torch.eq(
-                    methan_bonds,
-                    torch.tensor(
+    assert (
+        torch.any(
+            torch.eq(
+                methan_bonds,
```
chrisiacovella, line 493: and updated here: `methane_bonds`

chrisiacovella:
So I guess the real issue here is when we are setting up the dataset in the DataModule, we  are unaware of the NNP model (which determines whether we require unique pairs).   

I think the solution here is probably the easiest and best, as it retains the modularity (output of the DataModule can be passed to any NNP).  The other way to handle this would be to pass the model to the DataModule to extra the unique_pairs_parameter (since this is fundamental to the NNP, I don't think we'd want anyone having to set this on their, as that could be problematic).

chrisiacovella (approved):

modelforge/tests/test_dataset.py
wiederm: Thanks! Corrected the typo.

PR #153, opened 2024-06-10
Generating the precomputed pairlist in parallel

wiederm:
## Description
The pairlist is pre-computed, which takes about 40 minutes for the ANI2x dataset (the largest dataset in the collection). We can use the DataLoader logic to prepare these in parallel on multiple CPUs. 

This is unfortunately not as simple as I initially anticipated, since we are also removing self energies in the same pass. I will expand this PR with more information as soon as I have a better understanding of the DataLoader logic. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -745,20 +745,19 @@ class DatasetFactory:
         return TorchDataset(data.numpy_data, data._property_names)
 
 
-from torch.multiprocessing import Pool  
+from torch.multiprocessing import Pool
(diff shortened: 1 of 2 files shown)


chrisiacovella:
As mentioned offline, I think we can use the data loader class to loop through the pair list calculation, rather than having to use multiprocessing.

wiederm:
This PR has been superseded by PR #161

PR #155, opened 2024-06-11
Adding strategies for multitask optimization

wiederm:
## Description

There is no obviously correct way to combine forces and energy errors. These errors have large differences in their magnitude. The naive approach to linearly combine their MSE `Loss_total = \alpha_energy * MSE(energy) + \alpha_force * MSE(force)` and use scaling factors for the energy and force loss to consider the different magnitudes has been used in the field before. It might be a reasonable starting point. Reading through the literature on this [topic](https://github.com/Manchery/awesome-multi-task-learning#loss--gradient-strategy) I believe there are better ways explored by the multi-task learning community.

I will start by implementing the following:
- naive approach to combine the losses with scaling factors
- analytically combine the gradients (instead of directly combining the losses)

## Todos
  - [x] Combine losses with scaling factors

## Notes
Since we are using PyTorch Lightning for training, the fact that we require forces (or, in general, gradients) during validation and test steps makes this approach a bit complicated. PyTorch Lightning does automatically remove gradients.

## Status
- [x] Ready to go


wiederm:
In the process of working on this PR a few things came up:
- since we need derivatives in the validation/test steps this PR is setting `inference_mode` to false in the `Trainer` and manually sets `require_grad` on the positions passed to the `validation|test_step` methods
- the combined force loss that I have implemented looks like this `torch.mean((rho1 * (E_true - E_predict)**2 + (rho2 / (3 * N_atoms)) * torch.sum(torch.norm(F_true - F_predict, dim=1) ** 2)`, with rho1 and rho2 as user set hyperparameters
- Two closely related but separate concepts were pretty confusing to me: PyTorch Metrics and the Loss function. PyTorch Metrics can calculate any metric on the predicted values (given the true values) and logs these to the command line or on disk. The loss function is what's used to calculate the gradient. Initially, I was using Metrics to formulate the loss function, but this is intended to calculate *per epoch* properties and not necessarily  *per batch* properties. I have now implemented two different loss classes, one for Forces and Energies and one for Energies only. The output of the loss function is always a squared error, This is then passed to the MSEMetric and RMSEMetric and accumulated there, on epoch end the average is taken (and the square root for the RMSE error).

devtools/conda-envs/test_env.yaml
```diff
@@ -17,34 +17,30 @@ dependencies:
...
   - torchani
   - versioneer
 
   # Docs
   - sphinx_rtd_theme
 
-  # pip installs
   - pip:
-      - schnetpack>=2.0.0
+      # - schnetpack>=2.0.0
```
chrisiacovella, line 39: As we discussed offline, we should probably just move the tests for the original implementation to a separate subdirectory (with its own environment) and not run these with CI, just run before releases.

chrisiacovella (commented):
Can you merge the main into this (as this will stop the Mac tests from failing) and then apply the change to the CI file for schnet.   I think we probably want to move any tests that look at original implementations to their own directory, with own environment file, that we should just test manually.

chrisiacovella (approved):

devtools/conda-envs/test_env.yaml
wiederm: agreed!

PR #158, opened 2024-06-12
test TensorNet RSF

MarshallYan:
## Description
Implementation of radial symmetry functions for TensorNet.

## Todos
  - [ ] Lower limit of cosine cutoff function
  - [ ] Match units (angstrom in TensorNet; nanometer in modelforge)
  - [ ] Test whether results match TensorNet implementation

## Questions
- [ ] Do we need the lower limit of cosine cutoff function (like TensorNet did)?
- [ ] Aside from ExpNormalSmearing (default), TensorNet also has GaussianSmearing, do we also want to add that in modelforge?

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/potential/tensornet.py
+++ b/modelforge/potential/tensornet.py
@@ -4,9 +4,12 @@ import torch
 from openff.units import unit
 
 from modelforge.potential.models import InputPreparation
-from modelforge.potential.models import BaseNetwork, CoreNetwork
+from modelforge.potential.models import BaseNetwork
+from modelforge.potential.models import CoreNetwork
+from modelforge.potential.utils import CosineCutoff
 from modelforge.potential.utils import TensorNetRadialSymmetryFunction
-from modelforge.potential.utils import NeuralNetworkData, NNPInput
+from modelforge.potential.utils import NeuralNetworkData
+from modelforge.potential.utils import NNPInput
 
 @dataclass
 class TensorNetNeuralNetworkData(NeuralNetworkData):
@@ -96,10 +99,17 @@ class TensorNetRepresentation(torch.nn.Module):
     def __init__(
         self,
         radial_max_distance,
-        radial_min_distanc,
+        radial_min_distance,
         number_of_radial_basis_functions,
     ):
-        pass
+        super().__init__()
+
+        self.cutoff_module = CosineCutff(radial_max_distance)
+        self.radial_symmetry_function = self._setup_radial_symmetry_functions(
+            radial_max_distance, 
+            radial_min_distance, 
+            number_of_radial_basis_functions
+        )
 
     def _setup_radial_symmetry_functions(
         self,
@@ -116,7 +126,10 @@ class TensorNetRepresentation(torch.nn.Module):
         return radial_symmetry_function
 
     def forward(self, data: TensorNetNeuralNetworkData):
-        pass
+        radial_feature_vector = self.radial_symmetry_functions(data.d_ij)
+        # cutoff
+        rcut_ij = self.cutoff_module(data.d_ij)
+        radial_feature_vector = radial_feature_vector * rcut_ij
 
 
 class TensorNetInteraction(torch.nn.Module):
--- a/modelforge/tests/test_tensornet.py
+++ b/modelforge/tests/test_tensornet.py
@@ -4,3 +4,43 @@ from modelforge.potential.tensornet import TensorNet
 def test_tensornet_init():
     net = TensorNet()
     assert net is not None
+
+
+def test_compare_radial_symmetry_features():
+    # Compare the TensorNet radial symmetry function
+    # to the output of the modelforge radial symmetry function
+
+    import torch
+    from openff.units import unit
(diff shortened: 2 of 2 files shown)


MarshallYan:
On my local Fedora machine, if I don't install the current modelforge with `pip install ./` in directory "/home/shuaiy/workdir/modelforge", `from modelforge.potential.utils import CosineCutoff` wouldn't work with error being:
`>>> from modelforge.potential.utils import CosineCutoff
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/shuaiy/workdir/modelforge/modelforge/__init__.py", line 7, in <module>
    from ._version import __version__
ModuleNotFoundError: No module named 'modelforge._version'`

However, after `pip install ./`, the error disappears.

Does this have anything to do with the error occurs in testing?

MarshallYan:
What is the error with environment installation on macOS?

ianmkenney:
@MarshallYan It looks to me that the `ray-core` build `osx-arm64/ray-core-2.12.0-py311h15aa61a_0` is marked as broken in the [package files](https://anaconda.org/conda-forge/ray-core/files), which is the build currently used in the last CI pass in main. Looking at the latest CI pass, it appears that the [environment was cached](https://github.com/choderalab/modelforge/actions/runs/9514952817/job/26228193170#step:4:46). I assume if the environment cache is removed in the main CI, you'd see the same error.

It might be related, but to get the tests passing locally on osx-arm64, I needed to use the PyPI builds for all ray components. That's probably a separate issue that probably needs more consideration from @wiederm.

modelforge/potential/tensornet.py
```diff
@@ -96,10 +99,17 @@ class TensorNetRepresentation(torch.nn.Module):
...
         self,
         radial_max_distance,
-        radial_min_distanc,
+        radial_min_distance,
         number_of_radial_basis_functions,
     ):
-        pass
+        super().__init__()
+
+        self.cutoff_module = CosineCutff(radial_max_distance)
```
wiederm, line 107: The reason for the error in gh action is a typo here --- it should be `CosineCutoff`

wiederm:
thanks @ianmkenney for figuring out this error! I am on it!

wiederm:
@MarshallYan , the issue you mentioned [here](https://github.com/choderalab/modelforge/pull/158#issuecomment-2167092102) is normal --- you need to install the package so that its namespace is exposed in your current working environment.

wiederm:
@MarshallYan , you mentioned the lower limit of the cosine cutoff function. Can you please expand on this? How does this impact the cutoff behavior? Does the `lower limit` define the beginning of the influence of the Cutoff function?

modelforge/tests/test_tensornet.py
```diff
@@ -1,6 +1,49 @@
...
+    from torchmdnet.models.utils import ExpNormalSmearing
+
+    from modelforge.potential.utils import CosineCutoff
+    from modelforge.potential.utils import TensorNetRadialSymmetryFunction
+
+    # generate a random list of distances, all < 5
+    d_ij = torch.rand(5, 1) * 5
+
+    # TensorNet constants
+    radial_cutoff = 5.0  # radial_cutoff
```
wiederm, line 26: can you add units here pleaee?

modelforge/tests/test_tensornet.py
```diff
@@ -1,6 +1,49 @@
...
+    r_mf = r_mf * rcut_ij
+    
+    rsf_tn = ExpNormalSmearing(
+        cutoff_lower=radial_start,
+        cutoff_upper=radial_cutoff,
+        num_rbf=radial_dist_divisions,
+    )
+    r_tn = rsf_tn(d_ij)
+
+    assert torch.allclose(r_mf, r_tn)
```
wiederm, line 48: that is a very usefull test! if this passes we can merge this PR in the tenornet PR. 
Before merging in main we will refactor and move the pieces to the dedicated modules.

MarshallYan:
> @MarshallYan , you mentioned the lower limit of the cosine cutoff function. Can you please expand on this? How does this impact the cutoff behavior? Does the `lower limit` define the beginning of the influence of the Cutoff function?

In TensorNet, if the lower limit of cutoff function is set to anything larger than 0:

`
if self.cutoff_lower > 0:
    cutoffs = 0.5 * (
        torch.cos(
            math.pi
            * (
                2
                * (distances - self.cutoff_lower)
                / (self.cutoff_upper - self.cutoff_lower)
                + 1.0
            )
        )
        + 1.0
    )
    # remove contributions below the cutoff radius
    cutoffs = cutoffs * (distances < self.cutoff_upper)
    cutoffs = cutoffs * (distances > self.cutoff_lower)
    return cutoffs
`

MarshallYan:
@wiederm I made the radial symmetry function work by carrying  angstrom as the unit of everything just as what TensorNet does. I can techniqually hide that by doing the unit transformation inside class functions but does that make sense? What is our way of using `openff.units`? Right now it seems to me that this is an extra step because I still need to `*10` or `/10`.

Issue #151, opened 2024-06-10
Removing self energies / calculating neighborlist in parallel


wiederm:
The neighbor list and self energy calculations are currently done in sequence for each data point. As it is now the calculation takes around 40 minutes for the ANI2x dataset (the largest dataset in our collection right now). While this doesn't have high priority (it is a one time cost), I think we can parallalize this calculation using the PyTorch multiprocessing library with low additional efforts.

chrisiacovella:
I think we can improve on this without going down the "traditional" parallel route. 

In the current approach, we lop over the conformers and calculate the pairs, one conformer at a time.  Once we get this pair info for a conformer, we basically just append this to a list that at the end is used to generate a large tensor (with a second tensor that tells us how to index into it for a given conformer). 

Since we are just creating a single large tensor at the end, we should be able to just do this all at once (or in manageable batch sizes), we just need to slightly modify the revised code I put in issue #141 and end up in PR #146.   that bit of code was just a modification of the original algorithm, but instead of creating two  n*(n-1) tensors for each batch and then eliminating pairs that are not on the same molecule, this code generates of those tensors takes into account the size of each conformer, so we never generate pairs with other molecules. If we simply change this to remove the offsetting (keeping each pair at 0-index basis), this will generate all the appropriate pairs.  Since the number of pairs per conformer something we can analytically compute based on number of atoms, we can also generate the appropriate indexing tensor. 

Doing it this way should speed things up because it will be one (or a few) large vector operations, rather than many small ones.

wiederm:
I think this has been addressed in PR #161

PR #161, opened 2024-06-14
Pair list enumeration speed improvements

chrisiacovella:
As mentioned in #151 the largest datasets start to bog down in the pre-calculation of the pair list.  

Switching to a numpy-based function that operates on batches provides substantial speed improvements. I do use the torch DataLoader, with shuffle=False, so the order will be preserved for constructing the pair tensor and indexing tensors.  

In the prior version of the code, processing spice114 took about 13 minutes to generate the pair list and remove self energies.   This code now takes about 70 seconds for both operations. 

Marcus reported this step took roughly 40 minutes for these steps with ani2x, it now is about 4 minutes for the removing of self-energies and about 3 minutes for the pair list pre-computation.

Based on some experimentation, it seems like a batch size of 500 and 4 workers, no memory pinning and no persistent workers gives the best performance, on my system, although with the exception of memory pinning, I didn't see big differences between when modifying the parameters. 

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -40,7 +40,7 @@ class Metadata:
     F: torch.Tensor = torch.tensor([], dtype=torch.float32)
 
     def to(
-            self, device: Optional[torch.device] = None, dtype: Optional[torch.dtype] = None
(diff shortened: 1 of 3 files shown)


wiederm (approved):
That looks great!

PR #188, opened 2024-07-09
Fix typing issues

ArnNag:
## Description
I fixed some of the type annotations as well as replaced a couple `__eq__ `operations with `torch.eq`

## Status
- [ ] Ready to go


wiederm (approved):
Thanks for these fixes!

PR #189, opened 2024-07-10
Added in wrappers for importing external, optional libraries.

chrisiacovella:
## Description
As mentioned in issue #18 , there are several optional dependencies in this package (jax/flax/pytorch2jax), as we can perform the training without them;  also, currently we are installing them via pip.  It looks like pytorch2jax is no longer being used, as we decided to use ivy instead (and it was that package that required flax...although I note there still is code that calls PyTorch to jax in the main branch).  As such, we should be able to just switch to the conda-forge jax  (this was initially causing some issues, but seems to have been resolved).  I'll note though, since jax is still not necessary to train, we may still want to set this as an optional package (along with ivy). 

Additionally, right now ray is installed via pip; this package is not strictly necessary but still likely to be used in production. The conda install is only available for linux/windows; I think we stuck with pip initially for the Mac OS support, but since we've split our dev yaml files into separate linux and Mac ones (and are largely skipping any training operations on Mac), we should be able to use the conda-forge version for the linux build. 

Regardless if these are from pip or conda, any optional package that are called, should tell a user how to install, rather than just throwing an import error.  Full install directions will of course be provided in the documentation, but providing a descriptive error message with directions at the moment of failure, will help make it quick and easy for a user to get the right packages installed.   

Basically, this bit of code follows the approach done in mdtraj (also copied and extended in mbuild), whereby, we call a simple function that puts the call to import lib in  try/except structure.  if successfully imported, we return the imported library; if not successful, we provide a descriptive message about the package and how to install it before throwing an import error.  If we haven't defined a message for a given package, a generic one will be used. 

usage is simple:

```
from modelforge.utils.io import import_

jax = import_('jax')

```

If jax were not installed, we'd get a message in the console like:

```
jax is a library for numerical computing that is designed for high-performance machine learning research.

jax can be installed via pip:

    pip install jax

```

To handle importing submodules directly, in most cases we would just do something like:

```
jnp = import_('jax').numpy
```

which is basically equivalent to:

```
from jax import numpy as jnp
```

The Chem package in rdkit does not show up as an attribute in the main package, so to import we need to use this syntax `Chem = import_('rdkit.Chem')`.   It seems to be a little bit of a case-by-case basis which syntax to use (we aren't going to be widely using this so I'm not too worried). 

This also includes a `check_import` function that will try to load a module, then delete it if successful.  If unsuccessful, an import error will be raised, same as above.  This could be useful for doing a check if key packages are installed, but outside of the context they are needed (like, checking if torch and other key libraries are installed as soon as the modelforge library is imported). 

```
from modelforge.utils.io import check_import

check_import("ray")
```

I do think we should probably make packages in the curation submodule optional, since users are not likely to ever have to run those.  This would avoid installing rdkit, sqlitedict, retry, qcportal,and  qcelemental into the environment and having to satisfy those dependencies.  Again, this is easy to document, and the addition of descriptive messages for these packages would be good.   Currently, I've added in the "import_" wrapper for these functions. 

This also addresses the unrelated issue #193.  Testing of the angular feature vector for ani was failing related to the use of unstable sorting in the preparation of inputs.  This adds documentation of the underlying causes as a caveat and changes the test to consider the sum of the vector, rather than element-wise comparison of the vector. 

## Status
- [ ] Ready to go


wiederm (approved):
This is a great way to deal with the pip dependencies! It is also very helpful that all of these dependencies don't impact the core functionality making this approach very useful!

PR #192, opened 2024-07-10
Add nondimensionalizing scale factor to SAKE

ArnNag:
## Description
SAKE uses distances directly. This means that, as currently implemented, a portion of a weight matrix implicitly has dimensions of the reciprocal of distance. To avoid this, I introduce a scale factor hyperparameter that the user must provide with OpenFF units. This hyperparameter is internally converted to its magnitude in nanometers, and the distances (in nanometer) are divided by this magnitude.

## Questions
- [ ] Does using `d_ij` directly as done [here](https://github.com/ArnNag/modelforge/blob/3f662b881779b70cfe206f3670aa4b5705cb166b/modelforge/potential/sake.py#L340) result in a discontinuity at the cutoff distance?

## Status
- [ ] Ready to go


wiederm:
is this ready for review?

wiederm (approved):
LGTM!

PR #194, opened 2024-07-11
Ref postprocessing and loss

wiederm:
## Description

This PR refactors the postprocessing and loss calculation. 

Postprocessing module
------------------------------------

The behavior of the postprocessing module is controlled using the potential.toml file. An example (showing only the relevant entries for this PR) is shown below:

```toml
[potential.postprocessing_parameter]
[potential.postprocessing_parameter.per_atom_energy]
normalize = true
from_atom_to_molecule_reduction = true
keep_per_atom_property = true
[potential.postprocessing_parameter.general_postprocessing_operations]
calculate_molecular_self_energy = true
```

Every property provided by the core network can be accessed using its name (e.g., `per_atom_energy`) and then defining operations that can be performed on it (e.g., normalization, reduction operation). An additional boolean can be set if the per_atom property should also be present in the output. There are also `general_postprocessing_operations` that don't operate directly on the output of the core network but can still be calculated.

This is implemented using nn.ModuleDict and nn.Sequential to chain sequences of operations together (e.g. scaling and reduction).

Loss  module
------------------------------------
The loss module is now very compact, and depending on whether loss is calculated from [per-atom](https://github.com/choderalab/modelforge/blob/a0dde3199cec1f30cb01c9491f30cb73ba3cacf3/modelforge/train/training.py#L17) or [per-molecule](https://github.com/choderalab/modelforge/blob/a0dde3199cec1f30cb01c9491f30cb73ba3cacf3/modelforge/train/training.py#L67) errors, different modules perform the calculation.

The initialization is controlled by the training.toml file
```toml
[training.training_parameter.loss_parameter]
loss_property = ['per_molecule_energy_error', 'per_atom_force_error']
[training.training_parameter.loss_parameter.weight]
per_molecule_energy_error = 0.999
per_atom_force_error = 0.001
```

and the loss class is as follows:

```python
class Loss(nn.Module):
    """
    Calculates the combined loss for requested properties.

    Attributes
    ----------
    loss_property : List[str]
        List of properties to include in the loss calculation.
    weight : Dict[str, float]
        Dictionary containing the weights for each property in the loss calculation.
    loss : nn.ModuleDict
        Module dictionary containing the loss functions for each property.
    """

    _SUPPORTED_PROPERTIES = ["per_molecule_energy_error", "per_atom_force_error"]

    def __init__(self, loss_porperty: List[str], weight: Dict[str, float]):
        """
        Initializes the Loss class.

        Parameters
        ----------
        loss_property : List[str]
            List of properties to include in the loss calculation.
        weight : Dict[str, float]
            Dictionary containing the weights for each property in the loss calculation.

        Raises
        ------
        NotImplementedError
            If an unsupported loss type is specified.
        """
        super().__init__()

        from torch.nn import ModuleDict

        self.loss_property = loss_porperty
        self.weight = weight

        self.loss = ModuleDict()

        for prop, w in weight.items():
            if prop in self._SUPPORTED_PROPERTIES:
                if prop == "force":
                    self.loss[prop] = PerAtomToPerMoleculeError()
                else:
                    self.loss[prop] = PerMoleculeError()
                self.register_buffer(prop, torch.tensor(w))
            else:
                raise NotImplementedError(f"Loss type {prop} not implemented.")

    def forward(self, predict_target: Dict[str, torch.Tensor], batch):
        """
        Calculates the combined loss for the specified properties.

        Parameters
        ----------
        predict_target : Dict[str, torch.Tensor]
            Dictionary containing predicted and true values for energy and force.
        batch : Any
            The batch data containing metadata and input information.

        Returns
        -------
        torch.Tensor
            The combined loss for the specified properties.
        """

        loss = torch.zeros_like(predict_target["E_true"])

        for prop in self.loss_property:
            loss += self.weight[prop] * self.loss[prop](
                predict_target[prop], predict_target[f"{prop}_true"], batch
            )

        return loss
```
Questions
------------------------------------
As @chrisiacovella and I have already discussed, I think it would be a good idea to use the suffix convention: `per_atom_${property}`  and `per_molecule_${property}`. This might require some refactoring for the dataloader.

## Todos
  - [x] Refactor Postprocessing module
  - [x]  Refactor Loss module
  - [x]  adopt tests
 
## Status
- [x] Ready to go


wiederm:
@chrisiacovella , when you review, could you please take an extra look at the loss and logging module in the `TrainingAdaptor`? I am not completely convinced that this is the best way to combine these.

modelforge/potential/models.py
```diff
@@ -686,124 +682,180 @@ def _input_checks(self, data: Union[NNPInput, NamedTuple]):
...
+        for property in postprocessing_parameter:
+            if property.lower() in self._SUPPORTED_PROPERTIES:
+                self._registered_properties.append(property.lower())
+            else:
+                raise ValueError(
+                    f"Property {property} is not supported. Supported properties are {self._SUPPORTED_PROPERTIES}"
+                )
+
+        # register operations
+        for property, operations in postprocessing_parameter.items():
```
chrisiacovella, line 758: since this is looping over what was defined in the toml file, not over `self._SUPPORTED_PROPERTIES` (and see if the property key exists in `self._SUPPORTED_PROPERTIES`), doesn't this mean the order things are defined will influence the order of operations?  I thought we had decided we wanted it to be order independent (where the order of operations is defined within the code). 

 Right now we have't necessarily defined any sort of chained operations (everything is independent) so I don't think it matters yet, but I think it would be good to change this loop around future proof things or alternatively change the loops starting on line 749 (where we loop over the input operations) or 847 (where we finally call them).

modelforge/potential/models.py
```diff
@@ -686,124 +682,180 @@ def _input_checks(self, data: Union[NNPInput, NamedTuple]):
...
 
-        self.readout_operation = ModuleList(work_to_be_done_per_property)
-        self.readout_prop = props
+                    if self.dataset_statistic is None:
+                        log.warning(
+                            "Dataset statistics are required to calculate the molecular self-energies but haven't been provided."
+                        )
+                    else:
+                        atomic_self_energies = self.dataset_statistic[
+                            "atomic_self_energies"
```
chrisiacovella, line 800: Again, not sure if we wanted this to follow a specific naming convention: "energy-atomic_self" (although that doesn't as cleanly fit into that schema).

modelforge/tests/data/training_defaults/default.toml
```diff
@@ -20,17 +20,17 @@ cooldown = 5
...
 interval = "epoch"
 
 [training.training_parameter.loss_parameter]
-loss_type = "EnergyAndForceLoss"
-include_force = true
-force_weight = 0.99
-energy_weight = 0.01
+loss_property = ['per_molecule_energy', 'per_atom_force'] # use .
+[training.training_parameter.loss_parameter.weight]
+per_molecule_energy = 0.999 #NOTE: reciproce units
```
chrisiacovella, line 29: minor spelling issue.

chrisiacovella, line 29:
```suggestion
per_molecule_energy = 0.999 #NOTE: reciprocal units
```

modelforge/dataset/utils.py
```diff
@@ -124,8 +124,8 @@ def calculate_mean_and_variance(
         online_estimator.update(E_scaled)
 
     stats = {
-        "E_i_mean": online_estimator.mean,
-        "E_i_stddev": online_estimator.stddev,
+        "per_atom_energy_mean": online_estimator.mean,
+        "per_atom_energy_stddev": online_estimator.stddev,
```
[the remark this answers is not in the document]
chrisiacovella: I think it is fine to be [per_atom/molecule]_[property], we should just probably make sure we are totally consistent.

chrisiacovella (changes requested):
A few minor things in terms of naming, but also need to change things up a little if we want to ensure the order of operations of the post processing. That is, order things are defined in the file doesn't influence the order they are calculated, where we have a clearly defined order in the code (which would be necessary for chained operations).  In general, I don't like the idea that order matters in an parameter input file (if we wanted order to matter it would need to be in a script).

modelforge/potential/models.py
```diff
@@ -686,124 +682,180 @@ def _input_checks(self, data: Union[NNPInput, NamedTuple]):
...
+        self.dataset_statistic = dataset_statistic
+
         self._initialize_postprocessing(
-            processing_operation,
-            readout_operation,
-            dataset_statistic,
+            postprocessing_parameter,
         )
 
+    def _get_mean_and_stddev_of_dataset(self) -> Tuple[float, float]:
```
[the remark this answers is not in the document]
wiederm: this is a good point, I will change that

modelforge/potential/models.py
wiederm: Thanks for noting this! I was planning to loop through `self._SUPPORTED_PROPERTIES` but didn't change this it seems. I will modify the code. In that way the order can't be modified at runtime.

wiederm: We are now looping over `self._SUPPORTED_PROPERTIES`! But I think, in general, the best way to mitigate this is to define a sequence of operations per property that has an order but doesn't interact with the next sequence of operations for the same or different property. This is something that we have to ensure in the module and can't be user defined.

ArnNag:
https://github.com/choderalab/modelforge/blob/fd1f11d0214e818bde057b8fb8965a726d6225ea/modelforge/train/training.py#L38

Wouldn't it be more efficient to take the dot product with itself rather than computing the norm, then squaring?

ArnNag:
https://github.com/choderalab/modelforge/blob/fd1f11d0214e818bde057b8fb8965a726d6225ea/modelforge/train/training.py#L11
This class computes squared error, but does not mention the error function in the docstring or name. Perhaps we should call it `SquaredError`?

wiederm:
> https://github.com/choderalab/modelforge/blob/fd1f11d0214e818bde057b8fb8965a726d6225ea/modelforge/train/training.py#L38
> 
> Would it be more efficient to take the dot product with itself rather than computing the norm, then squaring?

That is a great suggestion! I believe your suggestion is more efficient (assuming the same cost per operation) since it avoids taking the square root. I have updated the line of code to this:
```(predicted_tensor - reference_tensor).pow(2).sum(dim=1, keepdim=True)```

PR #196, opened 2024-07-15
Remove pyg dependency

wiederm:
## Description
We have used `scatter_sum` instead of `Tensor.scatter_add_`. The `scatter_sum` operation from `pyg` performs tensor initialization to store the sum, broadcasting and other operations (setting dtype, device) under the hood. In some cases, this can lead to unexpected behavior. The main reason for the move to the native scatter operations is the need to install a separate package, `torch_scatter,`, which leads to a larger dependencies graph.

## Todos
  - [x] remove `torch_scatter` operations
  - [x] move to native Tensor.scatter_*
  
## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/doc_env.yaml
+++ b/devtools/conda-envs/doc_env.yaml
@@ -2,7 +2,6 @@ name: test
 channels:
   - conda-forge
   - pytorch
-  - pyg
 dependencies:
   # Base depends
   - python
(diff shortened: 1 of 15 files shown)


chrisiacovella:
Would it aid code clarity to provide a similar wrapper to scatter_sum?

chrisiacovella (approved):

PR #198, opened 2024-07-15
Make naming of "total_charge" and "positions" consistent between Prop…

ArnNag:
…ertyNames and NNPInput

## Description
It's a bit confusing to have different names for these fields when they represent the same thing.

## Status
- [ ] Ready to go


wiederm (approved):
That looks good! I think we will need to do this also for `E` and `F`, using `per_molecule_energy` and `per_atom_force`. But the place for the change to `F` and `E` is better split off from this PR since we haven't decided on an ontology for these names yet.

PR #203, opened 2024-07-19
fix inconsistency in weight/bias initialization

wiederm:
## Description
As @ArnNag noticed, the `Dense` implementation was inconsistent in how weights and biases are initialized vs how these are reset. During initialization and resetting, the implementation used different distributions to draw values.

## Todos
  - [x] call `reset_parameter` in `init`

## Questions
- [ ] I'd prefer to use the same activation function in each model. Since we are normalizing the atomic energies anyway, I don't think there is an issue with `nn.SiLu`. However, we need to empirically validate that this doesn't violate assumptions in some model architectures.

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/potential/utils.py
+++ b/modelforge/potential/utils.py
@@ -253,14 +253,15 @@ class Dense(nn.Linear):
         bias_init : Callable, optional
             Function to initialize the bias. Default is zeros_.
         """
-
-        super().__init__(in_features, out_features, bias)
-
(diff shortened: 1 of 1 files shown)


wiederm:
@ArnNag , I just realized that `reset_parameters` is called in the `__init__` of `nn.Linear`. We don't need to call this explicitly in the child class. I am merging this PR, but I only updated the docstring and added comments, making sure that we don't forget this again (no change to code).

PR #204, opened 2024-07-19
allow additional features in embedding

wiederm:
## Description

This PR adds a new feature that will allow to embed additional per-atom (e.g. partial charge, spin) and per-system properties (e.g. total charge). The approach is similar to what is outlined in this [PR](https://github.com/torchmd/torchmd-net/pull/289). 

The idea is to construct a feature vector containing nuclear charges and appending other per-system or per-atom properties, mixing these properties in a linear layer that will return a tensor that is then passed to the embedding layer. This doesn't require any change to the current architecture of the models.

## Todos
  - [x] add toml config parsing
  - [x] construct feature vector

## Status
- [x] Ready to go


modelforge/potential/sake.py
```diff
@@ -125,20 +126,20 @@ def _model_specific_input_preparation(
 
         number_of_atoms = data.atomic_numbers.shape[0]
 
-        atomic_embedding = self.embedding(
-            F.one_hot(data.atomic_numbers.long(), num_classes=self.max_Z).to(
-                self.embedding.weight.dtype
-            )
-        )
+        # atomic_embedding = self.embedding(
```
wiederm, line 129: @ArnNag , I am a bit surprised about the embedding here. Why are you not directly passing `data.atomic_numbers`? What is the expected output?

wiederm: I commented out the original code for now, but I'd like to understand this

Issue #200, opened 2024-07-17
Add data class for toml files


chrisiacovella:
Currently .toml files are read into dictionaries.  To improve upon this, let us load these into a data class (more specifically use pydantic to create a data model).   I think it makes sense that this data class will actually be 4 separate data classes ( dataset,, potential, training, runtime) to match the breakdown in the .toml files (and allow us to maintain that hierarchy of what goes with what).

wiederm:
that's a great idea!

Some thoughts on this: 
I would guess that at least every model needs its own dataclass. It seems difficult to come up with a superset of  all possible parameters for the potentials. A possible benefit of a superset would be that we can standardize naming conventions for the potentials.

I think the runtime toml file should only be used to override values set in one of the other toml files (a possible scenario where something like that is useful is when you have a single property that should differ between otherwise similar training runs, instead of defining a number of toml files that differ only in a single entire, we define a number of runtime toml files that include only the property that is different). Following from that use case I'd suggest that there shouldn't be a runtime dataclass (since it is only used to override fields).

wiederm:
I think it would be super useful to provide a useful serialization of the dataclass for logging purpose. It might be as simple as converting the fully initialized dataclass to a dictionary and `pretty_print`

wiederm:
Alternatively, runtime could also purely contain the entries that don't impact the training but the interaction with hardware and storage. That would mean the runtime.toml includes:
```
[runtime]
nr_of_epochs = 500
accelerator = "gpu"
num_nodes = 1
devices = 1                  #[0,1,2,3]
local_cache_dir = "./cache"
save_dir = "only_energy"
experiment_name = "{model_name}_{dataset_name}"
```

chrisiacovella:
I'll note, that initially I was implementing default values into the pydantic models.  Based on discussion today, we do not want them in there, so that there is no possibility for ambiguity.  If a parameter is necessary and not defined, the code will fail.   

For some functions it don't think it would hurt to have a default, parameters, e.g., like a logging frequency, as that should not impact code function, but probably better to not include that for consistency. 

There are a few parameters that likely do not need to be defined for training to work.  For example  "early stopping" and "stochastic weight averaging" for sure can be optional.   The learning rate scheduling also likely could be optional as well.

chrisiacovella:
[...]

PR #206, opened 2024-07-23
Major refactoring to use pydantic models for config files.

chrisiacovella:
## Description
This PR is related to issue #200, whereby .toml files (that hold parameters) are loaded into pydantic models, to streamline the code, enable validation, and make it easier for parameters to be documented than relying upon dictionaries internally.  

While we only need a single model to handle all the datasets, each potential requires their own model, however, the basic structure is the same. 

E.g. the CoreParameter section is more or less unique for each NNP, although  the post processing can use the same submodules (just with different defaults). 

```
class ANI2xParameters(ParametersBase):
    class CoreParameter(ParametersBase):
        angle_sections: int = 4
        radial_max_distance: str = "5.1 angstrom"
        radial_min_distance: str = "0.8 angstrom"
        number_of_radial_basis_functions: int = 16
        angular_max_distance: str = "3.5 angstrom"
        angular_min_distance: str = "0.8 angstrom"
        angular_dist_divisions: int = 8

    class PostProcessingParameter(ParametersBase):
        per_atom_energy: PerAtomEnergy = PerAtomEnergy(
            normalize=True,
            from_atom_to_molecule_reduction=True,
            keep_per_atom_property=True,
        )
        general_post_processing_operation: GeneralPostProcessingOperation = (
            GeneralPostProcessingOperation(
                calculate_molecular_self_energy=False,
                calculate_atomic_self_energy=False,
            )
        )

    model_name: str = "ANI2x"
    core_parameter: CoreParameter
    postprocessing_parameter: PostProcessingParameter
```

After several iterations, I think it is clearest to give each NNP parameter model a unique name (but using the same scheme, e.g, ANI2xParameters, SchNetParameters), and have these all stored in the same python file (potentials/parameters.py).  I also decided to nest the CoreParameter class inside the main parameter class for each, allowing us to retain the same internal naming structure, regardless of the NNP (since they are general, the PerAtomEnergy sub model, and GeneralPostProcessingOperation models live outside of the class structure). 

## Todo
- [x] Add in CI tests for the models themselves
- [x] Add in CI tests for training using the new models
- [x] Merge in main

## Status
- [x] Ready to go


chrisiacovella:
Currently running into issues on the CI, as qcportal and qcelemental do not seem to import properly.  This only started once I explicitly put in pydantic 2 in the environment file. qcportal uses pydantic 1, but using the .v1 namespace so it should be compatible with pydantic 2, but doesn't seem to be working as expected.   This does not appear to occur locally for me on either Mac or linux, and all installed packages seem to be the same, but it is failing on Mac and linux CI. 

I may just need to mark the tests to skip for now; the curation module is largely independent from the rest of the code (relies on some functions in utils), and the main training code does not depend upon it (also does not use pydantic directly).

chrisiacovella:
> Currently running into issues on the CI, as qcportal and qcelemental do not seem to import properly. This only started once I explicitly put in pydantic 2 in the environment file. qcportal uses pydantic 1, but using the .v1 namespace so it should be compatible with pydantic 2, but doesn't seem to be working as expected. This does not appear to occur locally for me on either Mac or linux, and all installed packages seem to be the same, but it is failing on Mac and linux CI.
> 
> I may just need to mark the tests to skip for now; the curation module is largely independent from the rest of the code (relies on some functions in utils), and the main training code does not depend upon it (also does not use pydantic directly).

The issue appears to have been a pin to an older version of qcelemental ( A version that apparently was using pydantic 1, not the .v1 namespace, so it was causing issues when pydantic 2 was installed).

chrisiacovella:
It's taken a lot of time to resolve test failures (seems like main was failing already when merged in), but I think I have gotten everything resolved.

wiederm:
We have used `NNP` and `potential` throughout the first versions. Since we are trying to standardize naming conventions, I am wondering if we should move to `model` instead?

modelforge/train/training.py
```diff
@@ -733,206 +733,178 @@ def configure_optimizers(self) -> Dict[str, Any]:
...
-    )
-    num_workers = dataset_config.get("number_of_worker", 4)
-    pin_memory = dataset_config.get("pin_memory", False)
-    local_cache_dir = runtime_config.get("local_cache_dir", "./")
-    # NOTE --------------------------------------- NOTE #
-    # FIXME TODO: move this to a dataclass and control default
-    # behavior from there this current approach is hacky and error prone
+    # log_training_arguments(
+    #     potential_config, training_config, dataset_config, runtime_config
+    # )
```
wiederm, line 882: I think we can remove this

modelforge/train/training.py
```diff
@@ -942,13 +914,22 @@ def perform_training(
...
+    )
+
     # Set up model
     model = NeuralNetworkPotentialFactory.generate_model(
         use="training",
         dataset_statistic=dataset_statistic,
-        model_parameter=potential_config,
-        training_parameter=training_config["training_parameter"],
+        model_parameter=potential_config.model_dump(),
+        training_parameter=training_config.model_dump(),
```
wiederm, line 932: I have a slight preference for `dict(potential_config)` instead of `.model_dum()` since it is more verbose of what is happening

modelforge/tests/test_schnet.py
```diff
@@ -166,24 +168,24 @@ def test_compare_forward():
...
+    config["potential"].core_parameter.number_of_radial_basis_functions = 5
+    config["potential"].core_parameter.number_of_filters = 12
 
     torch.manual_seed(1234)
 
     # initialize model
     schnet = SchNet(
-        **config["potential"]["core_parameter"],
-        postprocessing_parameter=config["potential"]["postprocessing_parameter"],
+        **config["potential"].model_dump()["core_parameter"],
```
wiederm, line 185: slight preference towards `dict(dataclass)`

chrisiacovella:
> We have used `NNP` and `potential` throughout the first versions. Since we are trying to standardize naming conventions, I am wondering if we should move to `model` instead?

I think `potential` is fine, as we've changed a lot of that.  I'll also note `model` is a protected namespace in pydantic, so it gets unhappy to have variables that start with `model_`

modelforge/train/parameters.py
```diff
@@ -0,0 +1,313 @@
...
+    cos = "cos"
+    linear = "linear"
+
+
+class Loggers(str, Enum):
+    """
+    Enum class for the experiment logger
+    """
+
+    wandb = "wandb"
```
wiederm, line 88: wandb needs a few additional parameters, how can we best add these?
```
        logger = WandbLogger(
            save_dir=save_dir,
            log_model=True,
            project=wandb_config["project_name"],
            group=wandb_config["group"],
            name=experiment_name,
            job_type=wandb_config["job_type"],
        )
```

wiederm (changes requested):
I think this is a great step towards input validation and read to merge in principle. I left a few comments that should be resolved first.

I saw that there are a few default parameters set tp `False`, do you think that there is a way to remove these?

wiederm:
> > We have used `NNP` and `potential` throughout the first versions. Since we are trying to standardize naming conventions, I am wondering if we should move to `model` instead?
> 
> I think `potential` is fine, as we've changed a lot of that. I'll also note `model` is a protected namespace in pydantic, so it gets unhappy to have variables that start with `model_`

That is a great point! Then let's use potential instead!

chrisiacovella:
> I think this is a great step towards input validation and read to merge in principle. I left a few comments that should be resolved first.
> 
> I saw that there are a few default parameters set tp `False`, do you think that there is a way to remove these?

I think I had put a note about this in the code.  This is only related to the post processing operations. 

 In order to define a model for the post processing operations, each possible operation variable needs to be defined in the model.   I chose to just mark all post processing operations to False; if you want to perform the post processing operations, you need to define it as True in the toml file. I think this fits with the general approach of no default parameters, because by default, if you do not define True, it does nothing.   

While these could be marked as "Optional", the code to parse this would still require us to check their value (the first check would be to see if not None, then checking the actual bool value).  Marking by default False in the model avoids this extra layer of checking and it will ultimately behave exactly the same.

modelforge/dataset/dataset.py
```diff
@@ -16,6 +16,47 @@
...
+    num_workers: int
+    pin_memory: bool
+
+    @field_validator("num_workers")
+    @classmethod
+    def check_num_workers(cls, v):
+        if v < 1:
+            raise ValueError("num_workers must be greater than or equal to 1")
+        return v
+
```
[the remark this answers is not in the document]
chrisiacovella: Good suggestion.  In Pydantic we can just use `num_workers: int= Field(gt=0)`. I had this as a user defined one because initially I had something in here to ensure that num_workers was less than the number of cores on the machine, but then realized that isn't actually necessary (it doesn't throw an error if you give it an absurd number of threads, like can sometimes happen with certain mpi implementations).

modelforge/dataset/dataset.py
```diff
@@ -1344,12 +1385,12 @@ def train_dataloader(
...
     ) -> DataLoader:
         """
-        Create a DataLoader for the training dataset.
+        Create a DataLoader for the runtime_defaults dataset.
 
         Returns
         -------
         DataLoader
-            DataLoader containing the training dataset.
+            DataLoader containing the runtime_defaults dataset.
```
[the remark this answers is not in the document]
chrisiacovella: ah, I think this is from merging.

modelforge/potential/parameters.py
```diff
@@ -0,0 +1,169 @@
...
+# To avoid having to set config parameters for each class,
+# we will just create a parent class for all the parameters classes.
+class ParametersBase(BaseModel):
+    model_config = ConfigDict(
+        use_enum_values=True, arbitrary_types_allowed=True, validate_assignment=True
+    )
+
+
+# define a reusable validator for converting string to unit.Quantity
+def convert_str_to_unit(value: Union[str, unit.Quantity]) -> unit.Quantity:
```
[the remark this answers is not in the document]
chrisiacovella: Great suggestion.  I renamed the function in units and am now just importing that.

modelforge/tests/test_schnet.py
chrisiacovella: dict(PydanticClass) does not recursively convert a nested model to a dictionary, only the top level.  In the case above, `dict(config['potential')['core_parameter']` would result in returning an instance of the CoreParameter pydantic class, not a dict of the contents. 

This is something I think we should actually discuss in terms of future road maps.  Do we want the functions to accept the pydantic models directly or maintain a pythonic interface of variables (or have a wrapper for the pythonic call that takes the pydantic model and does appropriate conversion).

modelforge/train/parameters.py
chrisiacovella: Since we only have two logger options at the moment, we can create a tensorboard class (with relevant parameters) and a wandb class.    I'm going to create and issue for this so we can do it in a separate, targeted PR.

chrisiacovella: I created issue #209 so we don't lose this comment

modelforge/train/training.py
chrisiacovella: good catch.

modelforge/train/training.py
chrisiacovella: I mentioned this in a comment above. model_dump() is required for a nested model.  As we just discussed in slack, I think after this release, we should work to moving towards an approach where all main entry points accept the pydantic models, rather than variables with defaults.  This will provide a much better means of validating inputs and tracking.

chrisiacovella:
I have everything addressed now, reading for a second look over @wiederm .

PR #207, opened 2024-07-23
Tensornet implementation

wiederm:
## Description
Implement tensornet using the parts that @MarshallYan has already built and implemented.

## Todos
  - [x] Setting up representation and embedding
  - [x] Implementing message creation
  - [x] Implement message passing

## Status
- [x] Ready to go


devtools/conda-envs/test_env.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - torchvision
   - openff-units
   - torchmetrics>=1.4
-  - pint
+  - pint=0.23
```
wiederm, line 20: is this pin necessary?

modelforge/potential/utils.py
```diff
@@ -1177,6 +1177,81 @@ def nondimensionalize_distances(self, distances: torch.Tensor) -> torch.Tensor:
...
+    ):
+        super().__init__(
+            number_of_radial_basis_functions,
+            max_distance,
+            min_distance,
+            alpha,
+            dtype,
+            trainable_centers_and_scale_factors,
+        )
+
```
wiederm, line 1198: You can remove this. As @ianmkenney mentioned, if you don't overwrite the __init__ method in the child class the __init__ of the super class is used

devtools/conda-envs/test_env.yaml
MarshallYan: This is necessary on my local setup.

modelforge/potential/utils.py
MarshallYan: Solved.

devtools/conda-envs/test_env.yaml
ArnNag: I was also having issues with openff-units and pint, not sure if that's related. I had to pin openff-units>=0.2.0

wiederm: we don't see this in CI, that's strange

wiederm: @chrisiacovella , have you seen this?

PR #208, opened 2024-07-29
Implement AIMNet2

ianmkenney:
This PR implements a version of the AimNet2 network published here: https://chemrxiv.org/engage/chemrxiv/article-details/662fbc9491aefa6ce1bebb89

THe documentation of the network in this publication is sparse, therefore some assumptions had to be made about the form of the calculations performed. Also, there is no mention of depths or width of any of the MLPs, so we are using reasonable default values here.

The network architecture is straightforward: it is an equivariant message-passing network that includes partial charge prediction (using multiple output heads, at least in this network implementation). At each message passing round, a message containing radial and vector features encoding distance-weighted embeddings/partial charges is constructed. Using an MLP, this message is used to iteratively refine the intial atom embeddings and partial charges. Partial charges are normalized to ensure that charge is preserved. The refined atom embedding is then used to predict per-atom energies, and the per-atom partial charges are used to predict long range electrostatics (not implemented yet).


modelforge/potential/aimnet2.py
```diff
@@ -168,54 +169,445 @@ def compute_properties(
...
+        radial_contributions = torch.zeros(
+            (per_atom_feature_tensor.shape[0], self.number_of_per_atom_features),
+            device=per_atom_feature_tensor.device,
+            dtype=per_atom_feature_tensor.dtype,
+        )  # Shape: (num_of_atoms, nr_of_features)
+
+        # Accumulate the radial contributions using index_add_
+        radial_contributions.index_add_(0, idx_j, proto_v_r_a)
+
+        # Step 2: Vector Contributions Calculation (Equation 5)
```
ianmkenney, line 289: This was the step we aren't completely certain about right? If so, a TODO comment would be a good visual indicator.

modelforge/potential/aimnet2.py
```diff
@@ -168,54 +169,445 @@ def compute_properties(
...
+        # Step 3: Compute the Euclidean Norm for each atom
+        vector_norms = torch.norm(
+            vector_contributions, p=2, dim=1
+        )  # Shape: (num_of_atoms,)
+
+        return radial_contributions, vector_norms
+
+
+class FirstMessageModule(BaseMessageModule):
+    def __init__(self, number_of_per_atom_features: int):
```
ianmkenney, line 322: This constructor calls the base class constructor which already calls:

```
self.linear_transform_embeddings = nn.Linear(
    number_of_per_atom_features, number_of_per_atom_features
)
```
You should be able to remove this.

However, seeing that the SubsequentMessageModule also does this, it looks like we're allocating resources to the `linear_transform_charges` tensor here when we shouldn't be.

modelforge/potential/aimnet2.py
```diff
@@ -168,54 +169,445 @@ def compute_properties(
...
+        feature_vector = torch.cat(
+            [radial_contributions_emb, vector_contributions_emb.unsqueeze(-1)], dim=1
+        )
+
+        return feature_vector
 
 
-class AimNet2Representation(nn.Module):
+class SubsequentMessageModule(BaseMessageModule):
+    def __init__(self, number_of_per_atom_features: int):
```
ianmkenney, line 386: Like the above constructor, this isn't doing anything that the base constructor isn't already doing---this can be removed. With both of these, it might just make more sense that base constructor is an abstract method.

modelforge/potential/aimnet2.py
```diff
@@ -168,54 +169,445 @@ def compute_properties(
...
+
+        return {
+            "per_atom_energy": E_i,
+            "per_atom_scalar_representation": atomic_embedding,
+            "atomic_subsystem_indices": data.atomic_subsystem_indices,
+        }
+
+
+class BaseMessageModule(nn.Module):
+    def __init__(self, number_of_per_atom_features: int):
```
ianmkenney, line 218: Consider making this an abstract method. See comments below on derived classes.

modelforge/potential/aimnet2.py
```diff
@@ -168,54 +169,445 @@ def compute_properties(
...
-        radial_symmetry_function = SchnetRadialSymmetryFunction(
+        radial_symmetry_function = SchnetRadialBasisFunction(
             number_of_radial_basis_functions=number_of_radial_basis_functions,
             max_distance=radial_cutoff,
             dtype=torch.float32,
         )
         return radial_symmetry_function
 
-    def forward(self, d_ij: torch.Tensor) -> Dict[str, torch.Tensor]:
+    def forward(self, data: Type[AIMNet2NeuralNetworkData]) -> Dict[str, torch.Tensor]:
```
ianmkenney, line 610: I think we should drop the `Type[...]` here. We aren't expecting the class itself as input, but rather an instance of `AIMNet2NeuralNetworkData`.

ianmkenney (commented):
Requesting changes (I can't make it since I'm technically the author of the PR).

Some minor things with class constructors and concerns over charge equilibration.

modelforge/potential/aimnet2.py
```diff
@@ -1,146 +1,150 @@
...
+        self.representation_module = AIMNet2Representation(
+            maximum_interaction_radius,
+            number_of_radial_basis_functions,
+            featurization_config=featurization_config,
+        )
+        number_of_per_atom_features = int(
+            featurization_config["number_of_per_atom_features"]
+        )
+
+        self.interaction_modules = torch.nn.ModuleList(
```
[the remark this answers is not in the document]
wiederm: great point! will modify to expose this

modelforge/potential/aimnet2.py
wiederm: will do!

modelforge/potential/aimnet2.py
```diff
@@ -1,146 +1,150 @@
...
+        The total charge of the system.
+    """
 
-        scale_factors = 0.5 / torch.square_(widths)
-        return scale_factors
+    total_charge = total_charge.unsqueeze(1)
+    total_charge = total_charge.expand_as(partial_point_charges)
+    partial_point_charges = partial_point_charges - torch.mean(partial_point_charges)
+    partial_point_charges = (
+        partial_point_charges * total_charge / torch.sum(partial_point_charges)
```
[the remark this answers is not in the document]
wiederm: great catch, I reworked this in PR #253

wiederm:
I moved all relevant pieces of code to PR #253 , and worked in the comments from @ianmkenney

PR #210, opened 2024-07-31
Smaller refactoring changes

wiederm:
## Description
This PR makes the following small naming changes:
- instead of `max_Z` for the highest element that defines the vocabulary for the embedding we use `highest_atomic_number`.
- instead of `cutoff` indicating the maximum interaction distance we use `maximum_interaction_radius`. This change is because some networks also define a minimum length, and the naming of that value was awkward, but `minimum_interaction_radius` is verbose and clear.

Additionally, to be more flexible, this PR includes changes allowing the user to define the activation function in the toml file. This clarifies what activation function is used and allows for rapid change. This won't work for all models out of the box; e.g., ANI uses the `CeLU` activation function with a defined alpha value. 

Some smaller changes include removing default values in some of the networks and adding missing docstrings.

## Question

@ArnNag , I noticed that SAKE uses 4 different activation functions. Are there empirical reasons for that, or can we use ShiftedSoftplus or SiLU for each case? 

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/potential/parameters.py
+++ b/modelforge/potential/parameters.py
@@ -2,10 +2,24 @@ from pydantic import BaseModel, field_validator, ConfigDict
 from openff.units import unit
 from typing import Union, List
 from modelforge.utils.units import _convert_str_to_unit
+from enum import Enum
 
 """
 This module contains pydantic models for storing the parameters of 
 """
+
+
+class ActivationFunction(str, Enum):
+    ReLU = "ReLU"
+    CeLU = "CeLU"
+    Sigmoid = "Sigmoid"
+    Softmax = "Softmax"
+    ShiftedSoftplus = "ShiftedSoftplus"
+    SiLU = "SiLU"
+    Tanh = "Tanh"
+    LeakyReLU = "LeakyReLU"
+    ELU = "ELU"
+
 
 
 # To avoid having to set config parameters for each class,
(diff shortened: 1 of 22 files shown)


modelforge/potential/parameters.py
```diff
@@ -32,18 +32,19 @@ class PerAtomEnergy(ParametersBase):
...
-        radial_min_distance: Union[str, unit.Quantity]
+        maximum_interaction_radius: Union[str, unit.Quantity]
+        minimum_interaction_radius: Union[str, unit.Quantity]
         number_of_radial_basis_functions: int
-        angular_max_distance: Union[str, unit.Quantity]
-        angular_min_distance: Union[str, unit.Quantity]
+        maximum_interaction_radius_for_angular_features: Union[str, unit.Quantity]
+        minimum_interaction_radius_for_angular_features: Union[str, unit.Quantity]
         angular_dist_divisions: int
+        activation_function: str
```
chrisiacovella, line 41: I think it would be best to define the activation function as an Enum rather than str. 

`class ActivationFunction(str, Enum):
    ReLU = "ReLU"
    CeLU = "CeLU"
    Sigmoid = "Sigmoid"
    Softmax = "Softmax"
    ShiftedSoftplus = "ShiftedSoftplus"
    SiLU = "SiLU"
    Tanh = "Tanh"
    LeakyReLU = "LeakyReLU"
    ELU = "ELU"
    `

chrisiacovella:
I like the change to being more explicit in the naming.  While very minor, for some reason "highest_atomic_number" seems strange to me.  Would it make more sense to stick with a consistent set of adjectives? For cutoff we are using minimum and maximum.  Previously this had been called "max_z", so maybe  "maximum_atomic_number"?

wiederm:
> I like the change to being more explicit in the naming. While very minor, for some reason "highest_atomic_number" seems strange to me. Would it make more sense to stick with a consistent set of adjectives? For cutoff we are using minimum and maximum. Previously this had been called "max_z", so maybe "maximum_atomic_number"?

That's the term that I was initially using in the refactoring but I changed it in the last commit. I am happy to use `maximum_atomic_number` though!

PR #212, opened 2024-08-01
fix bug in loading from single toml

wiederm:
## Description
Single toml file loading has been broken, this PR fixes the function.

## Status
- [x] Ready to go

### DIFF
--- a/modelforge/tests/test_training.py
+++ b/modelforge/tests/test_training.py
@@ -95,6 +95,17 @@ def test_train_with_lightning(potential_name, dataset_name):
     )
 
 
+def test_train_from_single_toml_file():
+    from modelforge.train.training import read_config_and_train
+    from modelforge.tests import data
+    from importlib import resources
+    import toml
+
+    config_path = resources.files(data) / f"config.toml"
+
+    read_config_and_train(config_path)
+
+
 import torch
 
 
--- a/modelforge/train/training.py
+++ b/modelforge/train/training.py
@@ -787,10 +787,10 @@ def read_config(
         runtime_config_dict = config["runtime"]
 
     else:
-        training_config_dict = toml.load(training_config_path)
-        dataset_config_dict = toml.load(dataset_config_path)
-        potential_config_dict = toml.load(potential_config_path)
-        runtime_config_dict = toml.load(runtime_config_path)
+        training_config_dict = toml.load(training_config_path)["training"]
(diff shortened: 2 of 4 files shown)


modelforge/train/training.py
```diff
@@ -787,26 +787,24 @@ def read_config(
...
 
     else:
-        training_config_dict = toml.load(training_config_path)
-        dataset_config_dict = toml.load(dataset_config_path)
-        potential_config_dict = toml.load(potential_config_path)
-        runtime_config_dict = toml.load(runtime_config_path)
+        training_config_dict = toml.load(training_config_path)["training"]
+        dataset_config_dict = toml.load(dataset_config_path)["dataset"]
+        potential_config_dict = toml.load(potential_config_path)["potential"]
+        runtime_config_dict = toml.load(runtime_config_path)["runtime"]
```
chrisiacovella, line 794: I think this is  probably better than having to reference the correct key when setting the parameters below.

modelforge/train/training.py
```diff
@@ -882,11 +880,11 @@ def read_config_and_train(
...
-        potential_parameters=potential_parameters,
-        training_parameters=training_parameters,
-        dataset_parameters=dataset_parameters,
-        runtime_parameters=runtime_parameters,
+    return perform_training(
+        potential_config=potential_config,
+        training_config=training_config,
+        dataset_config=dataset_config,
+        runtime_config=runtime_config,
     )
```
chrisiacovella, line 888: I also fixed this in the PR I'm working on.  I changed names for some reason when I decided to split this into a separate function (we have no testing of the entire function call which might be problematic).

chrisiacovella (approved):

modelforge/train/training.py
wiederm: I added a test in this PR for this particular function!

modelforge/tests/test_training.py
```diff
@@ -95,6 +95,17 @@ def test_train_with_lightning(potential_name, dataset_name):
...
 
+def test_train_from_single_toml_file():
+    from modelforge.train.training import read_config_and_train
+    from modelforge.tests import data
+    from importlib import resources
+    import toml
+
+    config_path = resources.files(data) / f"config.toml"
+
+    read_config_and_train(config_path)
```
wiederm, line 106: `read_config_and_train` calls `perform_training` @chrisiacovella

PR #218, opened 2024-08-06
Pydantic models for potentials can now accept an activation function and any associated parameters

chrisiacovella:
Pedantic models for the potentials will now accept the activation function and any associated parameters. These will automatically be checked and used to return a callable class with said parameters passed (if applicable). The potential init  function will get the function stored in a dictionary (because that is the form after dumping the pydantic model); all other internal function in the code expect a callable class instance.  I changed the name in the potentials from 
`activation_function` to  `activation_function_class` to make it a bit clearer that these are not names (there was some inconsistency in the codes). 

The fact this had to get passed as a dict I think highlights the importance of switching to  accepting pydantic models directly, rather than dumping the output.  It will certainly simplify the entry points to a lot of these functions, but that is not an insignificant undertaking and can be done after we get a stable release out. 

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/potential/parameters.py
+++ b/modelforge/potential/parameters.py
@@ -1,8 +1,21 @@
-from pydantic import BaseModel, field_validator, ConfigDict
+from __future__ import annotations
+
+from pydantic import (
+    BaseModel,
+    field_validator,
+    ConfigDict,
+    model_validator,
+    computed_field,
+)
 from openff.units import unit
-from typing import Union, List
+from typing import Union, List, Optional, Type
 from modelforge.utils.units import _convert_str_to_unit
 from enum import Enum
+
+import torch
+
+
+# needed to typecast to torch.nn.Module
 
 """
 This module contains pydantic models for storing the parameters of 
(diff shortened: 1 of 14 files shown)


modelforge/potential/utils.py
```diff
@@ -558,7 +560,7 @@ def forward(self, input: torch.Tensor):
 
         """
         y = F.linear(input, self.weight, self.bias)
-        return self.activation_function(y)
+        return self.activation_function_class(y)
```
wiederm, line 563: This might be difficult to understand, and I believe it is inconsistent with `nn.Identity()`.

We provide the class name, but then we initialize an instance and use it in the layers. We can also use the callable version since activity functions don't have an internal state. But here `nn.Identity()` is an instance, but the variable name implies it is a class.

wiederm: I think this is the better solution! Thanks!

modelforge/potential/parameters.py
```diff
@@ -18,7 +31,24 @@ def _missing_(cls, value):
         return super()._missing_(value)
 
 
-class ActivationFunction(CaseInsensitiveEnum):
```
wiederm, line 21: I really like the `CaseInsensitiveEnum` class!

modelforge/potential/physnet.py
```diff
@@ -665,8 +665,9 @@ class PhysNet(BaseNetwork):
         Number of interaction residual blocks.
     number_of_modules : int
         Number of PhysNet modules.
-    activation_function : str
-        Name of the activation function to use.
+    activation_function_parameter : Dict
+        Dict that contains keys: activation_function_name [str], activation_function_arguments [Dict],
+        and activation_function [Type[torch.nn.Module]].
```
wiederm, line 670: Oh, I was under the impression that the pydantic model takes the activation_function_paramter dict and returns an instance of the activation function?

wiederm (approved):
Thanks for this PR, that is super useful! 

I am wondering if we should put the initialization in the Pydantic module though?

modelforge/potential/physnet.py
chrisiacovella: It will take in the name and any possible parameters, and then set an instance of the class (with appropriate parameters) to a variable "activation_function"; since the pydantic model is hierarchical, the "activation_function" variable ends up a level down ( `core_parameter.activation_function_parameter.activation_function`).  Right now, the way the code is structured, we are dumping the pydantic model to a dict and then passing that to the NNP class using  argument unpacking (rather than parsing the model and passing individual arguments), rather than passing the pydantic model itself.  So having this accept the dict seemed like the fastest/least disruptive to the rest of the code, way to handle this.  

I think ultimately we absolutely should be passing the pydantic models instead, but I think we should wait until after the first release is cut.

PR #229, opened 2024-08-13
Fix command line argument

wiederm:
## Description
There were two separate bugs affecting command-line arguments:
- the device can be an `int` or a `List[int]`, this was not correctly handled by the `argparse` type check
- command line arguments should **override** parameters provided in the TOML file. The `read_config` function was doing the opposite (the docstring described the correct behavior)

## Status
- [x] Ready to go


chrisiacovella:
This looks good.  Good catch on parsing the devices;  I had this initially as just an int when I checked, then realized it could be a list and didn't properly test apparently (incorrectly assuming it would properly coerce the input).   I  pushed a minor change that looks for the presence of a comma instead of square brackets, so the input can be either "[0,1]" or "(0,1)" or "0,1"

wiederm:
Unfortunately, I think that won't work! It's kind of counterintuitive how this parameter works: when you pass an `int`, it means grab that number of GPUs, if you pass a `List[int]`, it means grab that specific GPU device. So, a common use case is to pass `[1]` to grab GPU:1 on a multi-GPU node. I think checking for `,` will miss that use case.

chrisiacovella:
> Unfortunately, I think that won't work! It's kind of counterintuitive how this parameter works: when you pass an `int`, it means grab that number of GPUs, if you pass a `List[int]`, it means grab that specific GPU device. So, a common use case is to pass `[1]` to grab GPU:1 on a multi-GPU node. I think checking for `,` will miss that use case.

Hmm, I'll switch it back, I guess I misunderstood what this is doing.  Maybe we need to eventually change the input syntax to make it clearer.  Like two separate arguments, `number_of_devices` (automatically picks) and `selected_devices` (you pick which ones). Even if internally we don't differentiate the functionality of those, it might be clearer to convey what is being done.

wiederm:
Yeah, this is confusing! THat wasn't my idea, but is what PyTorch Lighning is expecting

chrisiacovella (approved):

PR #230, opened 2024-08-14
Refactor downloader

chrisiacovella:
Previously I had created specific helper functions for downloading from zenodo and figshare. In particular, the figshare downloader made calls to their API to get the checksum and length of the file automatically.  I think this function was timing out sometimes during tests on the CI, causing failure (which is quite annoying when nothing about this function has been modified).  I revamped the code that relied upon these functions to use the more general `download_from_url` function that I had added in.  This required modifying the yaml files as well to have additional details (length of the file, not necessary but helpful for showing the progress of a download, and the name of the file to save the download to).  The yaml files already had the checksums encoded.   This should also make life easier; early in the project zenodo change their API breaking all the code and tests; we are no longer relying on the user apis from either platform. 

Tests have been revamped and these older, unnecessary functions have been completely removed.   

## Status
- [x] Ready to go


chrisiacovella:
Since this basically just swapped one existing function for another, I'm going to just skip review and merge.

PR #231, opened 2024-08-15
Added Black lint check GitHub Action

ianmkenney:
Since the contributor guide indicates we are using the Black formatting style, having a workflow that checks the format of the project source will make reviewing easier.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -15,5 +15,5 @@ jobs:
       - uses: actions/checkout@v4
       - uses: psf/black@stable
         with:
-          options: "--check --verbose"
+          options: "--check --verbose --line-length 88"
           src: "./modelforge"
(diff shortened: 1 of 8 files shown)


.github/workflows/black.yml
```diff
@@ -15,5 +15,5 @@ jobs:
       - uses: actions/checkout@v4
       - uses: psf/black@stable
         with:
-          options: "--check --verbose"
+          options: "--check --verbose --line-length 88"
```
chrisiacovella, line 18: this is great.  Can you put a note in the `docs/for_developer.rst` file that "We use Black's default line length of 88 characters."?

chrisiacovella (approved):

.github/workflows/black.yml
ianmkenney: Done!

PR #235, opened 2024-08-18
Long range electrostatics

wiederm:
## Description
This PR implements the infrastructure to calculate long-range electrostatics. In the first version, we will calculate the long-range term using a damped coulomb potential (see [ref](https://pubs.acs.org/doi/10.1021/acs.jctc.9b00181) for details). 

## Todos
  - [x] Add infrastructure to support the calculation of long-range electrostatic in a postprocessing step
  - [x] implement damped coulomb potential
  - [ ] add tests 

## Status
- [ ] Ready to go

### DIFF
--- a/modelforge/potential/parameters.py
+++ b/modelforge/potential/parameters.py
@@ -146,6 +146,7 @@ class ANI2xParameters(ParametersBase):
         minimum_interaction_radius_for_angular_features: Union[str, unit.Quantity]
         angular_dist_divisions: int
         activation_function_parameter: ActivationFunctionConfig
+        predicted_properties: List[PredictedPropertiesParameter]
 
         converted_units = field_validator(
             "maximum_interaction_radius",
@@ -275,7 +276,7 @@ class PhysNetParameters(ParametersBase):
         featurization: Featurization
         activation_function_parameter: ActivationFunctionConfig
         predicted_properties: List[PredictedPropertiesParameter]
-        
+
         converted_units = field_validator("maximum_interaction_radius")(
             _convert_str_to_unit
         )
--- a/modelforge/potential/parameters.py
+++ b/modelforge/potential/parameters.py
@@ -136,10 +136,20 @@ class PerAtomEnergy(ParametersBase):
     keep_per_atom_property: bool = False
 
 
+class CoulomPotential(ParametersBase):
+    electrostatic_strategy: str = "coulomb"
+    maximum_interaction_radius: Union[str, unit.Quantity]
+    
+    converted_units = field_validator(
+        "maximum_interaction_radius",
+    )(_convert_str_to_unit)
+
+
 class PerAtomCharge(ParametersBase):
     conserve: bool = False
-    strategy: str = "default"
+    conserve_strategy: str = "default"
     keep_per_atom_property: bool = False
+    coulomb_potential: Optional[CoulomPotential] = None
 
 
 class ANI2xParameters(ParametersBase):
--- a/modelforge/potential/parameters.py
+++ b/modelforge/potential/parameters.py
@@ -136,6 +136,12 @@ class PerAtomEnergy(ParametersBase):
     keep_per_atom_property: bool = False
 
 
+class PerAtomCharge(ParametersBase):
+    conserve: bool = False
+    strategy: str = "default"
+    keep_per_atom_property: bool = False
+
+
 class ANI2xParameters(ParametersBase):
     class CoreParameter(ParametersBase):
         angle_sections: int
@@ -157,6 +163,7 @@ class ANI2xParameters(ParametersBase):
 
     class PostProcessingParameter(ParametersBase):
         per_atom_energy: PerAtomEnergy = PerAtomEnergy()
+        per_atom_charge: PerAtomCharge = PerAtomCharge()
         general_postprocessing_operation: GeneralPostProcessingOperation = (
             GeneralPostProcessingOperation()
         )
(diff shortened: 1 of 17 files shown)


modelforge/potential/processing.py
```diff
@@ -334,3 +421,92 @@ def forward(self, data: Dict[str, torch.Tensor]) -> torch.Tensor:
...
+
+
+class LongRangeElectrostaticEnergy(torch.nn.Module):
+    def __init__(self, strategy: str, cutoff: unit.Quantity):
+        """
+        Computes the long-range electrostatic energy for a molecular system
+        based on predicted partial charges and pairwise distances between atoms.
+
+        The implementation follows the methodology described in the PhysNet
+        paper, using a cutoff function to handle long-range interactions.
```
chrisiacovella, line 433: Probably good to note that that this uses a cosine cutoff function, not the original function in physnet.

modelforge/potential/processing.py
```diff
@@ -334,3 +421,92 @@ def forward(self, data: Dict[str, torch.Tensor]) -> torch.Tensor:
...
+        phi_2r = self.cutoff_function(2 * pairwise_distances)
+        chi_r = phi_2r * (1 / torch.sqrt(pairwise_distances**2 + 1)) + (
+            1 - phi_2r
+        ) * (1 / pairwise_distances)
+
+        # Compute the Coulomb interaction term
+        coulomb_interactions = (
+            (per_atom_charge[idx_i] * per_atom_charge[idx_j])
+            * chi_r
+            / pairwise_distances
```
chrisiacovella, line 498: I don't think this equation is correct.  chi already takes into account the 1/r.  Basically, there is an extra  1/r (i.e., 1/pairwise_distances).  This is also missing coulombs constant which is 1.44 in eV*nm, so, this should be 138.96 in kj/mol nm (with e^2 in there).

modelforge/tests/test_models.py
```diff
@@ -26,7 +155,7 @@ def load_configs_into_pydantic_models(potential_name: str, dataset_name: str):
     dataset_config_dict = toml.load(dataset_path)
     potential_config_dict = toml.load(potential_path)
     runtime_config_dict = toml.load(runtime_path)
-
+    print(potential_config_dict)
```
chrisiacovella, line 158: probably should remove this.

chrisiacovella (changes requested):
The coulomb equation is not implemented correctly.  it has an extra 1/r and doesn't have the coulomb constant.

wiederm:
Thanks for looking through this. I believe this PR was not ready for review; tests are also missing. I will get this prepared until Monday.

modelforge/potential/processing.py
wiederm: great catch!

modelforge/potential/parameters.py
```diff
@@ -136,10 +130,22 @@ class PerAtomEnergy(ParametersBase):
...
 
 
+class CoulomPotential(ParametersBase):
+    electrostatic_strategy: str = "coulomb"
+    maximum_interaction_radius: Union[str, unit.Quantity]
+    from_atom_to_molecule_reduction: bool = False
+    keep_per_atom_property: bool = False
+
+    converted_units = field_validator(
+        "maximum_interaction_radius",
```
chrisiacovella, line 140: I think this should be the maximum_coulomb_interaction_radius (added in after the multiple pair listsPR). This might need some extra logic I. There that states like “use_local_interaction_pairlist” if cutoffs match.

wiederm: thanks! I changed the variable name

modelforge/potential/processing.py
```diff
@@ -449,10 +450,10 @@ def __init__(self, strategy: str, cutoff: unit.Quantity):
             The cutoff function applied to the pairwise distances.
         """
         super().__init__()
-        from .utils import CosineAttenuationFunction
+        from .utils import PhysNetAttenuationFunction
 
         self.strategy = strategy
-        self.cutoff_function = CosineAttenuationFunction(cutoff)
+        self.cutoff_function = PhysNetAttenuationFunction(cutoff)
```
chrisiacovella, line 456: Does it make sense to have the cutoff function be something the user could set (if the idea is to be able to eventually use this function in a more modular context)? The cosine function should still avoid a singularity at r_ij=0 (and the functions are still very closely related).   I'm not saying we necessarily have to make this user modifiable now.

modelforge/potential/processing.py
```diff
@@ -429,3 +419,90 @@ def forward(self, data: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
...
+        phi_2r = self.cutoff_function(2 * pairwise_distances)
+        chi_r = phi_2r * (1 / torch.sqrt(pairwise_distances**2 + 1)) + (
+            1 - phi_2r
+        ) * (1 / pairwise_distances)
+
+        # Compute the Coulomb interaction term
+        coulomb_interactions = (per_atom_charge[idx_i] * per_atom_charge[idx_j]) * chi_r
+        # 138.96 in kj/mol nm
+        # Zero out diagonal terms (self-interaction)
+        mask = torch.eye(
```
wiederm, line 499: that's also not right, it creates the squared matrix between all itneractions, not between all atoms.

modelforge/potential/processing.py
wiederm: that has been resolved

modelforge/potential/processing.py
chrisiacovella: Discussion offline: This might be something we want to be able to use this in a more general sense (although right now, the goal is to be able to implement physnet).

chrisiacovella (approved):

PR #241, opened 2024-08-21
remove pyproject.toml

wiederm:
## Description

This [PR](https://github.com/choderalab/modelforge/commit/09395a4ba0df739570ecc33cdc96d62707187912) added a `setup.py` file to manage the `setuptools` packaging. However, this seems to conflict with the packaging information in the `pyproject.toml` file. As a result, for the past few months, the correct dataset yaml files haven't been installed. This means that in my runs, I've been using the old datasets from Dropbox. This bug only became apparent when trying to train on SPICE, as the checksum in the yaml file didn't work with the hdf file on Dropbox.

One easy fix is to remove `pyproject.toml`. 

## Todos
  - [ ] Add test to make sure that we are downloading from zenodo

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


chrisiacovella:
Ian switched this up in PR #242 which is now merged.  I think we can remove this PR.

PR #242, opened 2024-08-21
Include  yaml+toml data files in MANIFEST.in and remove setup.py

ianmkenney:
YAML files in the `dataset` module were not included in the installation. Including them as an entry in the MANIFEST.in fixes this.


chrisiacovella:
These are included  in setup.py so they show up in the install.  Is the reason they need to be a manifest As well?

ianmkenney:
I was going based on comments from @wiederm and my local install not showing them in my environment. I think I wasn't seeing them because of an editable installation. The [data files support docs](https://setuptools.pypa.io/en/latest/userguide/datafiles.html) indicates you still need it though, so I'm a little unsure of what the absolute requirements are. It might be more important for `pyproject.toml` based builds.

ianmkenney:
`setuptools` pulls information from the `pyproject.toml`. So the `setup.py`, in its current form, isn't needed. If we ever have a more complicated build process that cannot be expressed in the `pyproject.toml` then it may be needed again. For the simple use of including data files, `pyproject.toml` + `MANIFEST.in` will do.

chrisiacovella (approved):
thanks for making this change!

PR #243, opened 2024-08-22
Refactoring changes in the training routine

wiederm:
## Description

This PR introduces a few improvements/changes to the training routine and its tests:
- improve logging messages
- improve fixture in conftest to obtain a single batch from different datasets and with different batchsize without hardcoding any of the values
- provide training toml file that sets parameters for force training
-  The Error classes will return now the error per molecule instead of the mean of the error, this makes it easier to log and use torchmetrics 
- tracking loss now also with torchmetrics instead of custom torch.nn.Module
- allowing to cache the processed dataset. In that case we assume that another process has already processed the dataset (since the `.pt` file is present) and, if cache regeneration is not requested, we will use this file and skiptt the `prepare_dataset` operation.
- lock the `prepare_dataset` method: only a single instance should execute this method per dataset.

This PR also includes bugfixes:
- fix a bug in the force loss calculation
- only retrain/create graph of force when in training mode (not in evaluation or test mode)

## Status
- [x] Ready to go


chrisiacovella:
The CI was failing due to running out of memory (CI runner is capped at 16 gb).  See #246 .  The code will skip training sake with forces.

modelforge/train/training.py
```diff
@@ -426,16 +436,22 @@ def _get_forces(
...
+
+        if grad is None:
+            raise RuntimeWarning("Force calculation did not return a gradient")
+
         per_atom_force_predict = -1 * grad  # Forces are the negative gradient of energy
 
         return {
             "per_atom_force_true": per_atom_force_true,
-            "per_atom_force_predict": per_atom_force_predict,
+            "per_atom_force_predict": per_atom_force_predict.contiguous(),
```
chrisiacovella, line 454: Why are you setting this to be contiguous here? I'm not seeing a this data used in a function that would require contiguous memory allocation but certain could be missing that.

chrisiacovella (approved):

scripts/config.toml
```diff
@@ -59,10 +57,9 @@ interval = "epoch"
 loss_property = ['per_molecule_energy', 'per_atom_force'] # use
 
 [training.loss_parameter.weight]
-per_molecule_energy = 0.999 #NOTE: reciprocal units
+per_molecule_energy = 0.009 #NOTE: reciprocal units
```
chrisiacovella, line 60: This has different weights than the training toml.  Is that intentional or was this just from experimentation?  I think it would probably be good to have the defaults be the same in the two files (even if we ultimately decide to change the defaults).

modelforge/train/training.py
wiederm: it is required by the Pytorch Lightning multi GPU training strategy, which makes assumptions about the memory layout of tensors (which might become non-contiguous if certain operations (view, permutate) are applied to the tensor.

scripts/config.toml
wiederm: good catch. Will fix this!

PR #247, opened 2024-08-31
Make it easier to test on single batch of dataset

wiederm:
## Description
Move the functions to test on a single batch of dataset from the test folder to the modelforge folder, makes it easier to use in notebooks/scripts and avoids duplication.

## Status
- [x] Ready to go


wiederm:
I will go ahead and merge this. The change only moves functions to new modules.

PR #248, opened 2024-09-02
Refactor for torchscript

wiederm:
## Description
As our project goals have evolved, we are now focusing on optimizing the models for TorchScript compatibility. This requires significant modifications and restructuring across the codebase, which will be addressed in this PR. The primary objective is to refactor the ModelFactory so that it assumes full responsibility for assembling a model. The model will consist of:
- a `CoreModel` (representing the specific architecture).
- a `Neighborlist` implementation (which will differ between training and inference).
- a `Postprocessing` module.

This involves the following changes:

### TorchScript Compatibility:

- Refactor the Neighborlist and Postprocessing modules to be TorchScript-compatible.
- Fix the dimensions of the Neighborlist and use masks to zero out contributions from atom pairs outside the cutoff radius.
- Eliminate inheritance patterns and the use of optional/union types, as these can complicate TorchScript compilation.

### Unit Handling:

- Remove all unit annotations and conversions from the core models, transferring the responsibility for unit conversions (to SI units) to the Pydantic models.

### Core Architectures:

- Modify each of the core architectures (with a potential challenge in the ANI architecture) to ensure TorchScript compatibility.
### Documentation and Refactoring:

Refactor docstrings and class structures to align with the new architecture and prepare for the upcoming release.

## Todos
 - [x] refactor the Neighborlist module to be compatible with TorchScript
 - [x] ensure that the Postprocessing module is TorchScriptable, including handling masks for atom pairs outside the cutoff
 - [x]  move unit conversion responsibilities to the Pydantic models, removing them from the core code
 - [x] refactor core architectures 
 - [x] testing
 - [x] prepare the codebase for a new release following the completion of all refactoring and updates

## Questions
- [x] Should we explore alternative implementations for the Neighborlist that might offer better performance or compatibility with TorchScript @chrisiacovella ?
- [x] Releated to the question above: if we want constant length neighborlist for a certain system, we will have atom pairs outside the cutoff in the neighborlist. We have cutoff functions that should get ride of the signal, but it might be necessary to zero these contributions out using a boolean mask when we aggregate the atom pairs for a central atom `i`. This  is something that we will have to check empirically. 

## Status
- [x] Ready to go


modelforge/tests/test_tensornet.py
```diff
@@ -1,122 +1,81 @@
...
+        simulation_environmen="PyTorch",
+        use_training_mode_neighborlist=True,
+    )
+
     from importlib import resources
     from modelforge.tests import data
 
+    # load reference data
     reference_data = resources.files(data) / "tensornet_input.pt"
+    reference_batch = resources.files(data) / "mf_input.pkl"
```
chrisiacovella, line 62: this file "mf_input.pkl" was not committed to the repo.

modelforge/potential/utils.py
```diff
@@ -359,18 +242,22 @@ class FeaturizeInput(nn.Module):
...
+        For per-atom categorical properties, we define an additional embedding
+        and add the embedding to the nuclear charge embedding.
 
         Parameters
         ----------
         featurization_config : dict
-            A dictionary containing the featurization configuration. It should have the following keys:
+            A dictionary containing the featurization configuration. It should
+            have the following keys:
             - "properties_to_featurize" : list
```
chrisiacovella, line 261: The structure of the toml seems to drop "properties_to_featurize".  I think we need to have this so that we know whether to throw an error if the properties about those features are not set.

wiederm: You are absolutely right; this has been addressed!

modelforge/potential/utils.py
```diff
@@ -442,7 +337,7 @@ def __init__(self, featurization_config: Dict[str, Union[List[str], int]]) -> No
                 int(featurization_config["number_of_per_atom_features"]),
```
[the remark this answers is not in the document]
wiederm: has been addressed!

wiederm:
@chrisiacovella , I am working on resolving the merge conflicts here. Please target this PR with the torchscriptable neighborlist.

wiederm:
@chrisiacovella , this PR is now ready for review!

modelforge/potential/__init__.py
```diff
@@ -2,29 +2,32 @@
...
-from .schnet import SchNet
-from .physnet import PhysNet
-from .painn import PaiNN
-from .ani import ANI2x
-from .sake import SAKE
-from .tensornet import TensorNet
+from enum import Enum
+
+from .ani import ANI2xCore
+from .aimnet2 import AimNet2Core
```
chrisiacovella, line 8: very minor, but can we group these imports with the other potential core inputs?

modelforge/potential/aimnet2.py
```diff
@@ -0,0 +1,750 @@
...
+            return torch.cat([feature_vector_emb, feature_vector_charge], dim=1)
+
+        # For the first message module, only return the atomic embedding contributions
+        feature_vector = torch.cat(
+            [radial_contributions_emb, vector_contributions_emb.unsqueeze(1)], dim=1
+        )
+        return feature_vector
+
+
+# class BaseMessageModule(nn.Module):
```
chrisiacovella, line 356: Can we remove this block of commented out code?

modelforge/potential/ani.py
```diff
@@ -657,113 +666,59 @@ def compute_properties(self, data: AniNeuralNetworkData) -> Dict[str, torch.Tens
...
-                minimum_interaction_radius_for_angular_features
-            ),
-            activation_function=activation_function,
-            angular_dist_divisions=angular_dist_divisions,
-            angle_sections=angle_sections,
-            predicted_properties=predicted_properties,
+    def _aggregate_results(
+        self, outputs: Dict[str, torch.Tensor]
+    ) -> Dict[str, torch.Tensor]:
+        per_atom_prediction = outputs.pop("per_atom_prediction")
```
chrisiacovella, line 685: I'm a little confused what is going on in this function.  Is this just basically just changing the hierarchy of the dict?  Can you add in some comments about what this does.

modelforge/potential/models.py
```diff
@@ -551,124 +546,114 @@ def apply_bwd(res, grads):
...
-            Parameters for configuring the training.
-        dataset_parameter : Optional[DatasetParameters], optional
-            Parameters for configuring the dataset.
-        dataset_statistic : Optional[Dict[str, float]], optional
-            Statistics of the dataset for normalization purposes.
-        potential_seed : Optional[int], optional
-            Seed for the random number generator.
-        simulation_environment : Optional[Literal["PyTorch", "JAX"]], optional, None
-            The simulation environment to use for training/inference. Will override the runtime parameter if provided.
+        cutoffs : Dict[str, unit.Quantity]
```
chrisiacovella, line 580: I think we need to update this doc string to be Dict[str, float].   I don't think the neighbor list needs units, as it would not be something a user would directly access.

modelforge/potential/models.py
```diff
@@ -551,124 +546,114 @@ def apply_bwd(res, grads):
...
-            else:
-                return model
-        else:
-            raise NotImplementedError(f"Unsupported 'use' value: {use}")
+
+            # Filter out the diagonal elements (where i == j)
+            mask = self.i_final_pairs != self.j_final_pairs
+            self.i_final_pairs = self.i_final_pairs[mask]
+            self.j_final_pairs = self.j_final_pairs[mask]
+        # calculate r_ij and d_ij
```
chrisiacovella, line 640: I'll make modifications in a separate PR after this is merged to speed things up (e.g., only using unique pairs and copying to get non-unique).  I'll note this takes in only_unique_pairs as an option, but doesn't do anything with it (but should).

modelforge/potential/models.py
chrisiacovella: Also I'm seeing the multiple pairlists have been removed from this PR. Is there a reason why?

modelforge/utils/units.py
```diff
@@ -43,6 +43,29 @@
 )
 
 
+def _convert_str_to_unit_length(val: Union[unit.Quantity, str]) -> float:
+    """Convert a string representation of an OpenFF unit to a unit.Quantity
+    If the input is already a unit.Quantity, it is returned as is.
```
chrisiacovella, line 48: Can you update the doc string to reflect this converts to nanometers and removes units?

modelforge/utils/units.py
```diff
@@ -43,6 +43,29 @@
...
+    Parameters
+    ----------
+    val : Union[unit.Quantity, str]
+        The value to convert to a unit.Quantity
+    Returns
+    -------
+    unit.Quantity
```
chrisiacovella, line 55: update to reflect it returns a float.

modelforge/potential/parameters.py
```diff
@@ -46,6 +46,16 @@ class ActivationFunctionParamsNegativeSlope(BaseModel):
     negative_slope: Optional[float] = None
 
 
+class AtomicNumber(BaseModel):
+    maximum_atomic_number: int = 101
```
chrisiacovella, line 50: I thought we had decided on not including default values in the models.

modelforge/potential/parameters.py
```diff
@@ -81,6 +81,30 @@ class ActivationFunctionParamsEnum(CaseInsensitiveEnum):
     ELU = ActivationFunctionParamsAlpha
 
 
+class PredictedPropertiesMixin:
+    predicted_properties: List[str]
+    predicted_dim: List[int]
+
+    # Custom validation for handling the predicted_properties dictionary
+    @model_validator(mode="before")
+    def handle_predicted_properties_dict(cls, values: dict) -> dict:
```
chrisiacovella, line 90: I don't think this routine is necessary at all. pydantic will automatically populate these fields.

modelforge/potential/parameters.py
```diff
@@ -81,6 +81,30 @@ class ActivationFunctionParamsEnum(CaseInsensitiveEnum):
     ELU = ActivationFunctionParamsAlpha
 
 
+class PredictedPropertiesMixin:
```
chrisiacovella, line 84: I don't think this should be a mixin here, as it isn't really necessary and I think just adds some complexity. 

The way I was handling these was to just create a separate pydantic class, and just set this as the type for the given parameter (e.g., see how ActivationFunction is set up). 

So we could just define

`class PredictedProperties(BaseModel):
    predicted_properties: List[str]
    predicted_dim: List[int]

    @model_validator(mode="after")
    def check_predicted_properties(self) -> PredictedProperties:
        if len(self.predicted_properties) != len(self.predicted_dim):
            raise ValueError(
                "The length of predicted_properties and predicted_dim must be the same"
            )

        return self`

And then in the each model, under CoreParameter, just add in `predicted: PredictedProperties`, e.g., like:


`class ANI2xParameters(ParametersBase):
    class CoreParameter(ParametersBase):
        angle_sections: int
        maximum_interaction_radius: Union[str, unit.Quantity]
        minimum_interaction_radius: Union[str, unit.Quantity]
        number_of_radial_basis_functions: int
        maximum_interaction_radius_for_angular_features: Union[str, unit.Quantity]
        minimum_interaction_radius_for_angular_features: Union[str, unit.Quantity]
        angular_dist_divisions: int
        activation_function_parameter: ActivationFunctionConfig
        predicted: PredictedProperties
`

The specific reason I did it this way was it keeps the model definitions reasonably self contained  which will make documentation easier.  We might also want to make the properties that are allowed to be predicted an enum, but we can do that later.

chrisiacovella (changes requested):
There are a few things to address.   To be able to train aimnet2 properly, I think we need to have the multiple cutoffs, but maybe I should address this in separate PR when we address the inference neighbor lists, and add back in all the support for the multiple cutoffs.

modelforge/potential/aimnet2.py
wiederm: removed the block

modelforge/potential/parameters.py
wiederm: Removed the default values. I will note that `maximum_atomic_number` does not impact the outcome of the calculations since it effectivly says 'what is the largest one-hot encoding that I will allow', which is in itself not immpacting downstream predictions. In other words, I think this is one of the things that we could set a default value for convencience.

modelforge/potential/parameters.py
wiederm: I didn't do a good job explaining what this does, but it is in fact necessary.

The input toml has the following structure

```
[potential.core_parameter.prediction]
predicted_properties = ["per_atom_energy"]
predicted_dim = [1]
```

that get's translated into a dictionary for which prediction is the key.

But the pydantic model wants `predicted_properties` and `predicted_dim` as keys (not inside a dictionary). To achieve this I need to 'flatten' the dictionary here.

modelforge/potential/parameters.py
wiederm: As pointed out in my last comment, I don't think this is a good idea. But I am happy to take any suggestions!

modelforge/potential/models.py
wiederm: I noticed that there is no effect if I set `only_unique_pairs` to `True`.

modelforge/potential/parameters.py
chrisiacovella: Would it just be easier to move this out of its own class at just put ‘predicted_properties’ and ‘predict_dims’ in CoreNetwork? We can still do model validation easily but the. They’d be properties the of core network and not prediction.

modelforge/potential/models.py
chrisiacovella: I can push a mini fix for this and a test. I fixed it in the nlist branch I started to feed into this. But rather fix this quickly in here otherwise ani will not work correctly in inference. Which would not be good.

modelforge/potential/ani.py
```diff
@@ -126,11 +107,74 @@ def __init__(
...
+    @staticmethod
+    def triple_by_molecule(
+        atom_pairs: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Input: indices for pairs of atoms that are close to each other.
+        each pair only appear once, i.e. only one of the pairs (1, 2) and
+        (2, 1) exists.
+
+        NOTE: this function is taken from https://github.com/aiqm/torchani/blob/17204c6dccf6210753bc8c0ca4c92278b60719c9/torchani/aev.py
+                with little modifications.
```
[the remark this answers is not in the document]
wiederm: Updated the docsting!

modelforge/potential/ani.py
wiederm: Updated the docstring

modelforge/potential/models.py
wiederm: thanks!

wiederm:
> There are a few things to address. To be able to train aimnet2 properly, I think we need to have the multiple cutoffs, but maybe I should address this in separate PR when we address the inference neighbor lists, and add back in all the support for the multiple cutoffs.

We can still train AimNet. When we predict partial charges, the loss will include terms for the dipole moment, which we will use to regularize/train the partial charge prediction. I agree that this is not necessarily the best way to train, but we can decouple the energy and partial charge prediction in such a way.

And I agree that it is best to address the neighborlist in a separate PR. The code was very complex because of the neighbor lists, and I removed them to move quickly on the torchscript implementation. But I agree that we should add this back in now.

modelforge/potential/ani.py
```diff
@@ -97,11 +86,11 @@ class ANIRepresentation(nn.Module):
...
         self,
-        radial_max_distance: unit.Quantity,
-        radial_min_distanc: unit.Quantity,
+        radial_max_distance: float,
+        radial_min_distance: float,
         number_of_radial_basis_functions: int,
-        maximum_interaction_radius_for_angular_features: unit.Quantity,
-        minimum_interaction_radius_for_angular_features: unit.Quantity,
+        maximum_interaction_radius_for_angular_features: float,
+        minimum_interaction_radius_for_angular_features: float,
```
chrisiacovella, line 93: Since the angular features will be set based on the pair list we calculated from the radial_max_distance, do we need to add in some logic in our parameters model validation to ensure that the angular_features cutoff is not larger than the radial_max_distance?

modelforge/potential/models.py
```diff
@@ -2,36 +2,43 @@
...
+T_NNP_Parameters = TypeVar(
+    "T_NNP_Parameters",
+    ANI2xParameters,
+    SAKEParameters,
+    SchNetParameters,
+    PhysNetParameters,
+    PaiNNParameters,
+    TensorNetParameters,
+    AimNet2Parameters,
+)
```
chrisiacovella, line 38: This will be useful!

chrisiacovella (approved):

modelforge/potential/ani.py
wiederm: that's a good point! we should check for this in the pydantic model

wiederm:
Thanks for the review @chrisiacovella ! I will go ahead and merge this!

PR #254, opened 2024-09-23
add dipole moment loss

wiederm:
## Description

Set up the infrastructure to use the dipole moment as a training target. 
This is, in principle, independent of an energy decomposition of the total energy; the loss operates only on the partial charges and the dipole moment. Using the shared network with property-specific output heads, each implemented potential can predict multiple scalar properties (per-atom energies, per-atom charges). In that way, it is reasonable to always include partial charge prediction in training when the dataset provides this property. 

The loss has the following form:

$$
L = | \sum_{i=1}^{N} q_i - Q^{\text{ref}}| +  \sum_{\alpha=1}^3 | \sum_{i=1}^N q_i r_{i, \alpha} - p_{\alpha}^{\text{ref}} |
$$

with
- $q_i$ as the predicted partial charges for atom $i$
- $Q^{\text{ref}}$ as the total charge of the molecule
- $r_{i, \alpha}$ as the $\alpha$ component of the coordinate of atom $i$
- $p_{\alpha}^{\text{ref}}$ as the dipol component $\alpha$

## Todos
  - [x] Expose dipole moment in the dataset
  - [x] Add loss for partial charge based on dipole calculation

## Status
- [x] Ready to go


modelforge/dataset/spice1.py
```diff
@@ -55,6 +55,7 @@ class SPICE1Dataset(HDF5Dataset):
         E="dft_total_energy",
         F="dft_total_force",
         total_charge="mbis_charges",
```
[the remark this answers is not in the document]
wiederm: Let me clarify: it seems as if we have passed 'mbis_charge' as `total_charge' to the training routine, yet 'mbis_charge' should be a partial charge scheme.

wiederm:
@chrisiacovella , I think we haven't exposed dipole moment for each of the datasets yet. I will do this in this PR, and I will add it as a property of the metadata (data that we need for training but shouldn't be passed to the network as input to avoid data leakage). 

In that process, I noticed that we have a partial_charge property in the `NNPInput` class; I will remove this.

wiederm:
@chrisiacovella , I might need you input  on that after all. 
I just realized that the `total_charge` record of the dataset (experimenting with SPICE1) is a property that exists per-record, which can be justified since the different conformations of a single record will have the same `total_charge`. Yet, this leads to the problem that when I naively populate the `property_of_interest["total_charge"]' tensor, I run out of indices when iterating over the different entries (since the indices goes over total_record * conformations.

Is there a good way to avoid this? For reference, the problem appears [here](https://github.com/choderalab/modelforge/blob/27469948f8bde4d2634249edb19f2b4dc10a31ee/modelforge/dataset/dataset.py#L310)

wiederm:
I just noticed that methods are initialized both in the `TrainingAdapter` and `ModelTrainer`. I will need a bit of time to resolve this, will fix in the next commit.

chrisiacovella:
So I think the issue is in the tag given to total charge.  It should be  `single_atom` I believe, but was `single_rec`.  Single rec was meant to differentiate things we wouldn't necessarily be training on, but might want in the dataset for filtering (like smiles or what data subset a record came from in spice, or just how many n_conf for a given system).   Definitely a mistake on my part which seems to have been copied/pasted across the spice/phalkethoh datasets.   I will update the curation classes as part of this, generate new versions, update yaml files, and post them to zenodo.

wiederm:
> Related, total_charge isn't necessarily what we want for the dipole correct? partial charge and dipole information is in the spice dataset already. do we need to update our data class to have "total_charge" and "charges" (as two separate pieces of information) or do we want to just use dipole directly?

I think the additional information that we want to
- predict partial charges
- use dipole moment in the loss functino
- calculate electronic energies

are 1) total charge and 2) dipole moment per data point (total charge won't change per conformation, but dipole moment will). I think we don't require partial charge from the dataset (we won't be using this for).

chrisiacovella:
@wiederm I'm going to fix it in this PR. My plan is to do a full revamp of datasets in a separate PR that will simplify and take into account a lot of the suggestions from over the summer in the discussions we had with Arnav.

chrisiacovella:
I've updated the spice dataset curation and datasets (all uploaded) such that total charge is now repeated as an array of shape (N_confs, 1).  This keeps current input validation in place.  It had a negligible impact on the size of the datafile to just repeat these values.

wiederm:
THanks @chrisiacovella !

I still get the error:
```2024-10-03 11:15:57.196 | WARNING  | modelforge.dataset.dataset:_metadata_validation:664 - Data keys used to generate dataset_test/data_test/qm9_test.json (['atomic_numbers', 'internal_energy_at_0K', 'geometry', 'charges']) do not match data loader (['geometry', 'atomic_numbers', 'internal_energy_at_0K', 'charges', 'dipole_moment']) ```

**Comment by chrisiacovella** [comment]
> Thanks for your help @chrisiacovella ! The last test that is failing is checking the shape of the predicted dipole moment vs the dataset dipole moment. I noticed that the shape of the SPICE2 dipole moment is (nr_of_molecules x 3), I would have expected (nr_of_molecules, 3). Should we reshape this at the dataset level? I guess most people are expecting a vector.

This isn't an issue with the datasets, but rather I think an issue with the collate function for the dipole moment. 

the output of the metadata for the energy is of shape (n_confs, 1), force is (sum_of_all_atoms, 3), and right now dipole moment is (n_confs*3) when it should be (n_confs, 3).  The issue is using torch.cat instead of torch stack.

**Comment by chrisiacovella** [comment]
I changed the shape of total_charge to be a 2d tensor to be consistent with other properties (e.g., energy); this is necessary to be able to sum up the total loss in a consistent manner (without unsqueezing for specific cases).  

Also added in some missing logic to convert names to the expected format (e.g., "dipole_moment" to "per_molecule_dipole"); also needed to set the true value in the CalculateProperties function for dipole moment.

**Comment by chrisiacovella** [comment]
I got all tests are passing; I will do a deeper dive into reviewing some more of the code tomorrow.

File: modelforge/train/training.py [hunk 2]
```diff
@@ -489,45 +146,105 @@ def _get_energies(
...
+    ) -> torch.Tensor:
+
+        per_atom_charge = model_predictions["per_atom_charge"]
+        # Compute predicted dipole moment
+        dipole_predict = torch.zeros(
+            (model_predictions["per_molecule_energy"].shape[0], 3),
+        ).scatter_add_(
+            dim=0,
+            index=batch.nnp_input.atomic_subsystem_indices.long().unsqueeze(-1),
+            src=per_atom_charge.unsqueeze(-1) * batch.nnp_input.positions,
```
chrisiacovella, line 193: This will only work if our molecule center of mass is at zero, which the datasets do not guarantee.  

We could do this as a preprocessing step when we load up the dataset (basically shifting coordinates such that COM is at the origin, as that likely will be easier than storing COM and then trying to do an operation on the entire batch.  This can likely just be done when we do the energy processing.

chrisiacovella (changes requested):

modelforge/train/training.py
chrisiacovella: I added in a preprocessing step to shift molecules to the center of mass, which happens basically at the same time we do the energy rescaling and pair list generation.  I'll note that even for spice2, this only took about 30 seconds, so we might want to consider doing this before training, even if we don't need the dipole moment in the loss function.  I did add in a validator in the training pydantic model that will throw an error if dipole moment appears in the loss function but shifting the center of mass is not set to True. 

Another minor change; at some point the code was switched from ignoring records (i.e., not loading that record)  if a record were missing a property of interest, to throwing an error.  I switched this back to having a warning in the log file.  It looks like  a handful of records in spice 2 (mostly seemed to be systems containing ions)  did not have the sci_dipole calculated (or it the array contained NAN).  I can look at that separately.

wiederm:
I think all has been addressed in this PR, will go ahead an merge

PR #257, opened 2024-09-26
Nlists for inference

chrisiacovella:
## Description
This  changes how we handle neighbor lists for inference.  This includes a more efficient brute force routine and a verlet scheme (although based still on an N^2 building step).  These changes will make it easier to support other libraries.

I included a version of the verlet scheme that builds the neighbor list using NNPOps;  however I commented it out right now to avoid adding that dependency as it does not seem to provide any improvements in speed/memory utilization after changing around the approach in the pure pytorch versions.  I think this can be made faster if we can push more of the required code pieces to the cuda kernel. 

box vectors are now stored in the NNPInput class to properly support periodic system.   

## Todo
- [x]  update inference neighbor list tests to point to new routines in neighbors.py. 

## Status
- [x] Ready to go


chrisiacovella:
I think this is ready to go.  I'm going to hold off on adding back in the multiple cutoffs and do that in a separate PR to keep this smaller and more focused.

modelforge/potential/neighbors.py
```diff
@@ -0,0 +1,418 @@
...
+
+        n = atomic_subsystem_indices.size(0)
+
+        # avoid reinitializing indices if they are already set and haven't changed
+        if self.indices.shape[0] != n:
+            # Generate a range of indices from 0 to n-1
+            self.indices = torch.arange(n, device=atomic_subsystem_indices.device)
+
+            # Create a meshgrid of indices
+            self.i_final_pairs, self.j_final_pairs = torch.meshgrid(
```
wiederm, line 138: does it make sense to downcast this to int32 and upcast when necessary (e.g. when using scatter operations)?

modelforge/potential/neighbors.py
```diff
@@ -0,0 +1,418 @@
...
+                self.indices, self.indices, indexing="ij"
+            )
+            # We will only consider unique pairs; for non-unique pairs we can just appropriately copy
+            # the data as it will be faster than extra computations.
+            mask = self.i_final_pairs < self.j_final_pairs
```
wiederm, line 143: I think I might be missing something here. `torch.meshgrid` creates the N^2 (btw, `torch.cartesian_prod` seems to do the same operation but returns it in the list view that we want. did you have a chance to look at this?), which, for our purpose, are all possible interactions (plus the self interaction on the diagonal). Would it not be easier not to remove the lower triangle and regenerate it subsequently, but simply mask the diagonal? 

The used mask:
````            mask = self.i_final_pairs < self.j_final_pairs```
can still be applied when only unique paris are required

File: modelforge/potential/neighbors.py [hunk 3]
```diff
@@ -0,0 +1,418 @@
...
+            )
+
+        else:
+
+            r_ij_full = torch.zeros(
+                total_pairs * 2, 3, dtype=positions.dtype, device=positions.device
+            )
+            d_ij_full = torch.zeros(
+                total_pairs * 2, 1, dtype=positions.dtype, device=positions.device
+            )
```
wiederm, line 182: I am surprised that this logic works.

My understanding is that e.g for a methane, using only unique paris, we have
```
pairs = [0,0,0,0,0,1,1,1,1,2,2,2,3,3,4],
             [1,2,3,4,5,2,3,4,5,3,4,5,4,5,5]
```
which reflects the upper triangle of the squared interaction matrix using index `i` and `j`.
The operation that you are suggesting is effectively:
- generate `d_ij` and `r_ij` for unique pairs
- duplicate and append to generate d_ij for all interactions

resulting in the following interactions
```
pairs = [0,0,0,0,0,1,1,1,1,2,2,2,3,3,4,0,0,0,0,0, ...],
        [1,2,3,4,5,2,3,4,5,3,4,5,4,5,5,1,2,3,4,5,...]
```
which doesn't look right.

But I might be missing something here!

modelforge/potential/neighbors.py
```diff
@@ -0,0 +1,418 @@
...
+
+            pairs = torch.zeros(
+                2, total_pairs * 2, dtype=torch.int64, device=positions.device
+            )
+
+            pairs[0][0:total_pairs] = temp1
+            pairs[1][0:total_pairs] = temp2
+            pairs[0][total_pairs : 2 * total_pairs] = temp2
+            pairs[1][total_pairs : 2 * total_pairs] = temp1
+
```
wiederm, line 413: can we extract this logic in a function?

and as I mentioned above, I think the special case should be `unique_pairs`, not non-unique pairs. IMO, when we use the meshgrid, non-unqie pairs are simply the flattened `grid_x` and `grid_y` outputs of the meshgrid, where the diagonal is removed.

wiederm (changes requested):

modelforge/potential/neighbors.py
wiederm: And, I guess this is something we only need to do once, right? we can keep the N^2 matrix of atom indices in memory?

modelforge/dataset/dataset.py
```diff
@@ -139,6 +144,7 @@ class NNPInput:
     total_charge: torch.Tensor
     pair_list: Optional[torch.Tensor] = None
     partial_charge: Optional[torch.Tensor] = None
+    box_vectors: Optional[torch.Tensor] = torch.tensor([])
```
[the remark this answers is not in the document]
chrisiacovella: It would be good to add that into the dataset setup.  I think we still need to think a bit about our "external" interface.  For openmm interface, it seems like if the system is not periodic, box vectors just aren't passed to the function call, so I guess we would know based on that what flags to set.

modelforge/potential/models.py
```diff
@@ -874,16 +728,14 @@ def setup_potential(
     potential_seed: Optional[int] = None,
     jit: bool = True,
     only_unique_pairs: bool = False,
-    box_vectors: Optional[torch.Tensor] = None,
     periodic: bool = False,
+    neighborlist_strategy: Optional[str] = None,
+    skin: Optional[float] = 0.08,
```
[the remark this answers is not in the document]
chrisiacovella: Optional does work in torch script, but it is a bit non-intuitive.  Probably better to remove this.

modelforge/potential/neighbors.py
chrisiacovella: I don't think that is necessary. this is not the bit that causes us to run out of memory as system size grows, and ultimately this will not be the routine we are going to use primarily (after we update NNPOps).

modelforge/potential/neighbors.py
chrisiacovella: We don't store the full N^2 matrix in memory after the initial calculation;  in the two lines below this we mask and save only the subset (which is the upper triangular removing the diagonal).  The Cartesian product basically doesn't the same thing as mesh grid but returns shape (N,2), not (2,N) which is what we want.  So I'm not sure what advantage this would be.

modelforge/potential/neighbors.py
chrisiacovella: We don't just concatenate, we flip indices for the second half. 


so given a 3 atom system with unique pairs with i !=j :

```
pairs = [[0,0,1], 
              [1,2,2]]
```

flipping the indices for the second half becomes:

```
pairs = [[0,0,1,1,2,2], 
             [1,2,2,0,0,1]]
```

Which includes all the relevant non-unique pairs.

modelforge/potential/neighbors.py
chrisiacovella: The reason why I have this only calculate for unique pairs, and then copy if we need non-unique pairs is efficiency.  The copy operation is negligible cost compare to doing twice as many computations (pretty much exactly a factor of 2).   It would probably be good to make this a separate function since it is used in both the brute and vert list.

modelforge/dataset/dataset.py
```diff
@@ -163,8 +169,11 @@ def to(
                 if self.partial_charge is not None
                 else self.partial_charge
             )
+            self.box_vectors = self.box_vectors.to(device)
+
         if dtype:
             self.positions = self.positions.to(dtype)
+            self.box_vectors = self.box_vectors.to(dtype)
         return self
```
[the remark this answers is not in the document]
chrisiacovella: good call.

modelforge/dataset/dataset.py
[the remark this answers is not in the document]
chrisiacovella: I refactored a bit after putting "is_periodic" into NNPInput.  periodicity is no longer passed to the constructor of the displacement function, but passed every time we call it (just like box vectors).  This should be more portable for the model.

modelforge/potential/neighbors.py
chrisiacovella: I'll note we only generate once and just save this, unless the number of particles changes (in which case we need to regenerate).

modelforge/dataset/dataset.py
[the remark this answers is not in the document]
wiederm: that's great!

modelforge/potential/models.py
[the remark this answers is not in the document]
wiederm: agreed!

modelforge/potential/neighbors.py
chrisiacovella: This is now in a separate function in each class to make it a little cleaner.  However, the same logic still applies of copying unique pairs data into non-unique pair data.

modelforge/potential/neighbors.py
wiederm: > We don't store the full N^2 matrix in memory after the initial calculation; in the two lines below this we mask and save only the subset (which is the upper triangular removing the diagonal).

Yes, I understand. But I was wondering if it's not better to store the N^2 matrix in memory. The upper triangular is only a special case needed for ANI, and can be easily recovered from the N^2 matrix. Therefore I am wondering if it is not easier and clearer to operate on the N^2 matri (and only removing the diagonal here)

modelforge/potential/neighbors.py
wiederm: sorry, my post is a bit missleading: I think what I mean is, that for a given index $i$, the indices $j$ are increasing.

modelforge/potential/neighbors.py
wiederm: oh, thank you for clarifying! Please ignore my post above in that case! That is a sufficient reason to be slightly non-intuitive :-)

modelforge/potential/neighbors.py
wiederm: you clarifyied your seasons below, I consider this as resolved!

modelforge/potential/neighbors.py
wiederm: I'd still prefer to use float32 and int32 wherever we can and upcast when necessary. Do you think that we might run in an overflow here that requires int64?

chrisiacovella: I think we will run out of memory before we run out of indices for this algorithm. However, changing these to int32 right now will cause a cascade of a downstream failures at this point.  This might be good to explore outside this PR, to see if it has any really impact on efficiency and memory utilization; we would definitely need to put in a bunch of upcasting and probably need to define some global variable to set the dtype for indexing.

wiederm: Thanks for looking into this! I agree though, let's keep it at int64

chrisiacovella:
I added in a basic doc page about the neighbor lists.  We will still need to think about the API for controlling things.  We might ultimately need a wrapper inside the setup model that allows us to change the neighbor list implementation, instead of it being something that we build the model with (if I built with it, we need to build different versions of the model with different neighbor lists).  But we can address that as we work on the openmm interface.

Issue #264, opened 2024-10-06
Potential.load_state_dict() in training mode not compatible with loading a training model


MarshallYan:
Key modification in Potential.load_state_dict() in training mode causing an error.

wiederm:
To clarify: the current workflow is intended for training a model in `training` mode, then saving the state dict and loading into a model that is created in `inference` mode. This is done because during training, the potential is wrapped in a training pytorch-lightning class that handles the boilerplate training code.

The issue you are describing appears when saving a state_dict in training mode and loading it in a new model also in training mode. The anticipated workflow for such a use case was a checkpoint file.
The issue is caused by the custom overloading of the `load_state_dict()` method.

wiederm:
I think this has been resolved

PR #265, opened 2024-10-06
Skip load_state key modification in training mode

MarshallYan:
### Associated Issue(s)
 - [x] Key modification in load_state_dict() in training mode causing an error #264.

## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/devtools/conda-envs/docs_env.yaml
+++ b/devtools/conda-envs/docs_env.yaml
@@ -40,3 +40,4 @@ dependencies:
       - pytorch2jax
       #- "ray[data,train,tune,serve]"
       - git+https://github.com/ArnNag/sake.git@nanometer
+      - torchviz2
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -37,7 +37,6 @@ dependencies:
       - flax
       - pytorch2jax
       - git+https://github.com/ArnNag/sake.git@nanometer
-      #- "ray[data,train,tune,serve]"
       - flax
       - torch
       - torchviz2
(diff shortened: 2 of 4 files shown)


wiederm:
@MarshallYan , I am not sure if this modification is necessary.

I reverted the modification that you made to the `load_state_dict` method and outlined the 3 use cases in this [test](https://github.com/choderalab/modelforge/blob/b81639f015f7d912f66338ec801694a923d38119/modelforge/tests/test_models.py#L343). They are currently all passing. Could you please first add the use case to this test and check that it is failing.

MarshallYan:
> @MarshallYan , I am not sure if this modification is necessary.
> 
> I reverted the modification that you made to the `load_state_dict` method and outlined the 3 use cases in this [test](https://github.com/choderalab/modelforge/blob/b81639f015f7d912f66338ec801694a923d38119/modelforge/tests/test_models.py#L343). They are currently all passing. Could you please first add the use case that you are imagining and that is failing to this test and check that it is failing.

I went through the test and it did seem to pass. However, `trainer.model` may not be used as expected because:
 1. `trainer` should be a `ModelTrainer` object
 2. `ModelTrainer.model` is initialized by `ModelTrainer.setup_potential()`
 3. `ModelTrainer.setup_potential()` returns a `TrainingAdapter` object
 4. So, `trainer.model` is actually a `TrainingAdapter` object

Now, in the tests you linked to, since `trainer.model` is a `TrainingAdapter` object, and model2 is a `Potential` object, they actually called different `load_state_dict()` methods. This is super confusing to me but I agree that this shouldn't be fixed here.

So, in conclusion, I think we can simply close this PR and directly work on seperate two methods for creating models for training and inference.

wiederm:
Merging this because it outlines the different use cases in the test.

PR #268, opened 2024-10-08
Optimize the training routine

wiederm:
## Pull Request Summary
Add techniques to train efficiently/effectivly:
- gradiet clipping
- adding tests for the different loss functions
 
 - [x] PR is ready for review


wiederm:
@chrisiacovella , this is a quick PR that adds tests for the different loss functions in training. It targets the #254 PR

modelforge/tests/test_training.py
```diff
@@ -80,36 +75,100 @@ def get_trainer(potential_name: str, dataset_name: str, training_toml: str):
...
 
+    if "force" in loss:
+        add_force_to_loss_parameter(config)
+    if "normalized" in loss:
+        replace_per_molecule_with_per_atom_loss(config)
+    if "dipole_moment" in loss:
+        add_dipole_moment_to_loss_parameter(config)
+
+    # train potential
+    get_trainer(config).train_potential().save_checkpoint("test.chp")  # save checkpoint
```
wiederm, line 169: the `get_trainer` function might be an idea instead of the factory class @MarshallYan @chrisiacovella

modelforge/potential/models.py
```diff
@@ -907,6 +907,7 @@ def setup_potential(
         jit=jit,
         jit_neighborlist=False if use_training_mode_neighborlist else True,
     )
+    model.train()
```
chrisiacovella, line 910: Since `setup_potential` is used to set up models for interference, why are you are marking it in training mode here?

modelforge/train/training.py
```diff
@@ -44,6 +44,14 @@
 ]
 
 
+def _exchange_per_atom_energy_to_per_molecule_energy(prop: str):
```
chrisiacovella, line 47: super minor here, but this should be either `_exchange_per_atom_energy_for_per_molecule_energy` or `_change_per_atom_energy_to_per_molecule_energy`

modelforge/train/training.py
```diff
@@ -183,15 +189,25 @@ def _predict_dipole_moment(
         self, model_predictions: Dict[str, torch.Tensor], batch: BatchData
     ) -> torch.Tensor:
 
-        per_atom_charge = model_predictions["per_atom_charge"]
+        per_atom_charge = model_predictions["per_atom_charge"]  # Shape: [num_atoms]
+        positions = batch.nnp_input.positions  # Shape: [num_atoms, 3]
+        per_atom_charge = per_atom_charge.unsqueeze(-1)  # Shape: [num_atoms, 1]
+        per_atom_dipole_contrib = per_atom_charge * positions  # Shape: [num_atoms, 3]
```
chrisiacovella, line 195: I think it would be good to add a comment that dipole moment is calculated by weighting charge by the position of the charge; this assumes that the center of mass of the system is at the origin.

modelforge/train/training.py
```diff
@@ -183,15 +189,25 @@ def _predict_dipole_moment(
...
+
+        indices = batch.nnp_input.atomic_subsystem_indices.long()  # Shape: [num_atoms]
+        indices = indices.unsqueeze(-1).expand(-1, 3)  # Shape: [num_atoms, 3]
+
```
chrisiacovella, line 199: Listing all the shapes is very useful!

modelforge/train/training.py
```diff
@@ -359,6 +375,7 @@ def _update_metrics(
         """
 
         for prop, metric_collection in metrics.items():
+            prop = _exchange_per_atom_energy_to_per_molecule_energy(prop)
```
chrisiacovella, line 378: it would be good to add a comment that this only modifies `per_atom_energy` and all other properties are left unchanged.

chrisiacovella (approved):
A few, mostly, minor things.

modelforge/potential/models.py
wiederm: good catch! that was definitly not intended, setting this to eval

modelforge/train/training.py
wiederm: thanks :-) adopted your first suggestion

modelforge/train/training.py
wiederm: I hope that we can do this by using custom tensor classes in the future. this would help a lot and would implicitly document shapes (especially (nr_atoms, 1), (nr_atom, 3), (n_pairs,1), (n_pairs, n_features) are common enough that this might make sense

modelforge/train/training.py
wiederm: added comments and docstrings

wiederm:
All comments have been addressed! Will merge this PR

PR #270, opened 2024-10-09
Set value to monitor during training in single place

wiederm:
## Pull Request Summary

Multiple Training Callbacks executed at certain points during the training process based on model performance. Model performance is measured based on a metric that has to be passed by name to the trainer. To avoid having to set the same metric name (e.g., `val/per_molecule_energy/rmse`) multiple times this PR changes the TOML structure and allows passing a single `monitor` key:value pair that is used if the key is not present in other callbacks (so there is still the possibility for each callback to monitor a different metric).

 - [x] PR is ready for review


wiederm:
@chrisiacovella , this is ready for review.

Issue #277, opened 2024-10-12
Speeding up CI


chrisiacovella:
CI is currently taking at least 30 minutes on the linux nodes. To try to speed this up, I think using pytest-xdist might be good, as it will let us run a lot of the faster tests in parallel.  The testing of the training of models using a lot of memory, so those should be run in serial (or at least those tests should not be run concurrently with each other).

chrisiacovella:
Looking at the numbers from the last commit to main:

- ubuntu 3.11: 44m 16s
- ubtuntu 3.10: 44m 23s
- macOS 3.11: 53m 30s
- macOS 3.10: 48m 42s

chrisiacovella:
I added a wiki page on this to provide more details, including how to use locally.   https://github.com/choderalab/modelforge/wiki/Developer-Notes#implementing-unit-tests-and-pytest


The summary: pytest-xdist allows parallel execution of tests.  It did not provide any improvement on the GitHub macOS runners, so it is only implemented for ubuntu runners in the CI.  

Because of memory concerns on the GitHub runners, we are mostly stuck with running the tests in `test_training.py` serially.  We can still run other tests concurrently while running the training tests, but that means are total elapsed time will be max(time to running test_training.py, time to run all other tests), which will be in the 20-25 minute range.   This is still much more reasonable to be able to see if there are errors than waiting up to 45 minutes. 

Also,  as noted in the benchmarks on the wiki, this can drastically speed up local testing on a machine with a decent number of cores and memory, further removing testing from being a bottleneck.

On my linux laptop (with 16 physical cores and 64 gb of ram), run all tests except the training tests (i.e., not including test_training.py):

- standard pytest : 27 min 49 s
- 1 thread: 22 min 8 s
- 2 threads: 13 min 34 s
- 4 threads: 7 min 57 s
- 8 threads: 4 min 48 s
- 16 threads : 3 min 26 s
- 32 threads : 2 min 58 s


only running test_training.py (note any more 8 threads and I run out of memory for the tests).
- original pytest: 18 min 40 sec
- 1 thread : 20 min 15 secs
- 2 threads: 10 min 53 sec
- 4 threads: 9 min 20 sec
- 8 threads : 9 min 17 sec 


Running 8 concurrent threads for all tests: 12 min and 38 sec.

PR #278, opened 2024-10-12
Speeding up CI using pytest-xdist

chrisiacovella:
## Pull Request Summary
The CI is taking quite a long time to run (about 45 minutes).  This aims to speed this up using pytest-xdist, which will allow concurrent execution of tests.  To do this though, we need to make sure our tests will not be in conflict (make sure any files being written out/read have unique names, making sure all unique test files are at minimum using their own unique temp directories). 

Edit: actual numbers from the last merge into main:

- ubuntu 3.11: 44m 16s
- ubtuntu 3.10: 44m 23s
- macOS 3.11: 53m 30s
- macOS 3.10: 48m 42s

### Associated Issue(s)
 - [x] #277

## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -78,7 +78,7 @@ jobs:
       - name: Run tests
         # conda setup requires this special shell
         run: |
-          pytest -n {matrix.nprocs} --dist loadgroup -v --cov=modelforge --cov-report=xml --color=yes --durations=50 modelforge/tests/
+          pytest -n {matrix.n_procs} --dist loadgroup -v --cov=modelforge --cov-report=xml --color=yes --durations=50 modelforge/tests/
 
       - name: CodeCov
         uses: codecov/codecov-action@v4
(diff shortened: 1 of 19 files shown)


chrisiacovella:
using `auto` which sets the number of concurrent process to the number of logical cores (which in this case I think is 2 for ubuntu):

- ubuntu 3.11: 23m 49s
- ubtuntu 3.10: 23m 8s
- macOS 3.11: 45m 32s
- macOS 3.10: 59m 22s

Clearly this is not effective on macOS, but that is okay, since we are primarily focused on the ubuntu tests, so as long as those can complete faster so we can get more immediate feedback that would be helpful.

chrisiacovella:
Setting -n 8 seemed to speed things up substantially, but ran out of memory as the test_training_with_lightning tests started to run; setting -n4 seems to make no difference compared to auto for ubuntu.  So will switch back to auto. Since it is slower on macOS using parallel, this will be set to 0 which is the default.  Again, the point is mostly to speed up ubuntu so we can get faster turn around time as to whether things are passing.

chrisiacovella:
Codecov randomly took 17 minutes on Ubuntu 3.10. Ignore that (usually takes <10 s) both ubuntu CI timings are about 24 minutes.

Issue #269, opened 2024-10-08
NNP Factory Class Function Redesign


MarshallYan:
Based on my discussion with @wiederm and @chrisiacovella, it might be beneficial to redesign the logic of this middle layer "factory" and rename anything that is currently named as "model".

Note: We should explicitly call a model a "trainer" when it is a `ModelTrainer` (the class itself should be `PotentialTrainer`) object in training mode, and a "potential" in inference mode when it is a `Potential` object.

The general idea is that generate a potential **trainer** or loading a potential **trainer** from either a checkpoint or a Pytorch state_dict should be an attribute of the `ModelTrainer` class. To load a potential for inference, on the other hand can be implemented by a) a new function in the `Potential` class, b) a new class named as `PotentialInferrer` which holds those JAX converter, etc, or c) a function directly placed in "models.py" (we should call it "potential.py" respectively). *I personally prefer the b) option.*

In this design:
  1. There won't be confusion of what a model is.
  2. All trainer properties are under its class, so when a user tries to initialize a trainer, there's no need to jump between "models.py" (will be "potential.py") and "training.py".
  3. Anything about inference will be in "potential.py" since you will use a **potential** model to do inference and should have nothing to do with "training.py".
  4. Upload model will be implemented under a Trainer class as there's no reason to upload models when they're used only for inference.
  5. Download models will be implemented seperately for training and inference. A user will have to choose to download a a) checkpoint file, b) a state_dict file, or c) an entire model file ("potential.pt", we need to further whether is necessary at any situation). This way, we can easily set a naming convention (or separate directories holding different types of potential models) so that users will only download inference potential for inference and vice versa.

chrisiacovella:
To elaborate a little more, this would involve removing the `NeuralNetworkPotentialFactory`. This class is basically just to switch between setting up a `ModelTrainer` instance for training, or return an instance of a `Potential` for inference.  We can instantiate the `ModelTrainer` directly from, e.g., `read_config_and_train`.  We would probably want another wrapper as Shuai mentioned, because we do have some extra logic needed related to exporting to pytorch or converting to jax. 

[...]

PR #283, opened 2024-10-15
Rename variable names in NNP factory class

MarshallYan:
## Pull Request Summary
Rename variables to distinguish among "Potential", "ModelTrainer", and "TrainingAdapter" objects in models.py, training.py, test_models.py, and helper_functions.py.

### Key changes
 - [x] "trainer": all `ModelTrainer` objects
 - [x] "potential": all `Potential` objects
 - [x] "potential_training_adapter": all `TrainingAdapter` objects
 - [x] "model": either a `ModelTrainer` object or a `Potential` object (will be refactored and separated in another PR)
 - [x] Rename "ModelTrainer" to "PotentialTrainer"

### Associated Issue(s)
 - [x] #269 

## Pull Request Checklist
 - [x] PR is ready for review


modelforge/potential/models.py
```diff
@@ -482,19 +482,20 @@ def setup_potential(
...
     )
-    model.eval()
-    return model
+    potential.eval()
+    return potential
 
 
 from openff.units import unit
 
+from modelforge.train.training import ModelTrainer
```
chrisiacovella, line 498: Were you going to change the name of `ModelTrainer` to `PotentialTrainer`?

modelforge/potential/models.py
```diff
@@ -520,7 +521,7 @@ def generate_potential(
         jit: bool = True,
         inference_neighborlist_strategy: str = "verlet",
         verlet_neighborlist_skin: Optional[float] = 0.1,
-    ) -> Union[Potential, JAXModel, pl.LightningModule]:
+    ) -> Union[Potential, JAXModel, pl.LightningModule, ModelTrainer]:
```
chrisiacovella, line 524: Same here, were we going to change this to PotentialTrainer?

chrisiacovella (changes requested):
We probably should rename "model.py" to "potential.py".   Are you using an IDE where it will automatically refactor?  If not I can do this and push a change.

chrisiacovella:
Lint is failing; might just need to check line lengths in model.py, training.py, and test_training.py

modelforge/potential/models.py
MarshallYan: That actually makes more sense, and I am changing it now.

MarshallYan:
> ## Pull Request Summary
> Rename variables to distinguish among "Potential", "ModelTrainer", and "TrainingAdapter" objects in models.py, training.py, test_models.py, and helper_functions.py.
> 
> ### Key changes
> Notable points that this PR has either accomplished or will accomplish.
> 
> * [ ]  Change 1
> 
> ### Associated Issue(s)
> * [ ]  Issue 1
> 
> ## Pull Request Checklist
> * [ ]  Issue(s) raised/addressed and linked
> * [ ]  Includes appropriate unit test(s)
> * [ ]  Appropriate docstring(s) added/updated
> * [ ]  Appropriate .rst doc file(s) added/updated
> * [ ]  PR is ready for review



> We probably should rename "model.py" to "potential.py". Are you using an IDE where it will automatically refactor? If not I can do this and push a change.



> We probably should rename "model.py" to "potential.py". Are you using an IDE where it will automatically refactor? If not I can do this and push a change.

Are you saying to rename the file and update with git? I can do that, or I can just wait until you push a change to refactor class names inside. For simplicity, could you push a commit of renaming files (if I understood correctly)?

MarshallYan:
> Lint is failing; might just need to check line lengths in model.py, training.py, and test_training.py

Yeah I will check. I did rename "self.model" to "self.potential_training_adapter" as they are actually hodling a TrainingAdapter objects.

chrisiacovella (changes requested):
Major change was renaming models.py to potential.py to be consistent with our recently agreed upon naming scheme.  While at it, I fixed some of the formatting issues in test_training, and fixed a few of the tests in test_potentials.py that I missed in terms of making sure all tests right to temp directories.  

Since the names have changed, the .rst files in the docs still need to be updated to reflect the naming change.  I think we can merge after the docs are updated.

modelforge/potential/potential.py
```diff
@@ -435,6 +426,18 @@ def setup_potential(
...
+            {
+                "training_dataset_statistics": {
+                    "per_atom_energy_mean": unit.Quantity(0.0, unit.kilojoule_per_mole),
+                    "per_atom_energy_stddev": unit.Quantity(
+                        1.0, unit.kilojoule_per_mole
+                    ),
+                }
+            },
+        )
+
```
wiederm, line 440: it might make sense to leave it in the signature of the function. it makes it very clear how a non-default dictionary should look like by presenting the default.

modelforge/train/training.py
```diff
@@ -665,7 +663,9 @@ def __init__(
             else dataset_statistic
         )
         self.experiment_logger = self.setup_logger()
-        self.model = self.setup_potential(potential_seed)
+        self.potential_training_adapter = self.setup_potential_training_adapter(
```
wiederm, line 666: `potential_training_adapter` is a descriptive name, but I am not sure if it is a useful name. It is long and it requires an understanding of the class structure of the training module.

maybe it makes more sense to call this lightning_module?

wiederm (approved):
Great changes @MarshallYan and @chrisiacovella !

chrisiacovella:
It looks like the only thing that still needs to be addressed is updating the .rst files in docs.

chrisiacovella (approved):
This looks good. Let's get this merged.

modelforge/train/training.py
MarshallYan: This completely make sense to me!

modelforge/potential/potential.py
MarshallYan: I would suggest not using any `Dict` as a default value of a function as it may have some unexpected behaviors when you call the function multiple times or you reimport the module (reload the function). The problem is that the object is created when loading the function https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python.

chrisiacovella: But we weren't passing an empty dictionary.  If `dataset_statistic` isn't passed, it will be initialized to the default values.  It's a bit confusing though, because these values can be overwritten based on the "use_default_dataset_statistic" flag, which if false by default, will read the dataset statistics from file (although I note there is no checking on whether this file exists).  I think this all needs a bit of a refactor in terms of the dataset_statistic  because there are way too many layers and branches.

chrisiacovella: Mostly, I don't think this should keep us from merging the PR; these changes in the PR are overall straightforward and thus should be easy to merge, but this will impact other PRs downstream.

Issue #293, opened 2024-10-21
Cannot save torch script models


chrisiacovella:
After the last merging of code, saving a potential that has been processed through torch.jit.script ends up with the error 

```
"RuntimeError: strides() called on an undefined Tensor"

```

The following code produces this error:

```
import torch
from modelforge.potential.potential import NeuralNetworkPotentialFactory
from modelforge.utils.misc import load_configs_into_pydantic_models

config = load_configs_into_pydantic_models("ani2x", "qm9")

potential = NeuralNetworkPotentialFactory.generate_potential(
    use="inference",
    potential_parameter=config["potential"],
    potential_seed=42,
)
print(type(potential))

potential_jit = torch.jit.script(potential)

from modelforge.dataset.dataset import single_batch

batch = single_batch(1, "qm9", "./")

print(potential_jit(batch.nnp_input))

potential_jit.save("ani2x_jit.pt")
```

I'm assuming this is related to the change to slots for NNPInput,

PR #295, opened 2024-10-22
Fixing NNPInput to allow us to write torch script models

chrisiacovella:
## Pull Request Summary
When changing NNPInput from a tuple to using slots, pair_list and per atom charge were set to None.  While we can call torch.jit.script on the potential with None in the NNPInput and evaluate it, we can't write it to file as it goes through all the validation.  

This defines these as empty tensors; we can check the shape of the pair_list in the Neighborlist to know if we need to generate these pairs. 

### Associated Issue(s)
 - [x] #293 
 
## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/modelforge/jax.py
+++ b/modelforge/jax.py
@@ -47,12 +47,4 @@ def convert_NNPInput_to_jax(nnp_input: NNPInput):
         nnp_input.per_atom_partial_charge
     )
 
-    # if nnp_input.pair_list is not None:
-    #     nnp_input.pair_list = convert_to_jax(nnp_input.pair_list)
-    #
-    # if nnp_input.per_atom_partial_charge is not None:
-    #     nnp_input.per_atom_partial_charge = convert_to_jax(
-    #         nnp_input.per_atom_partial_charge
-    #     )
-
     return nnp_input
--- a/modelforge/jax.py
+++ b/modelforge/jax.py
@@ -1,4 +1,4 @@
-from modelforge.dataset import NNPInput
+from modelforge.utils.prop import NNPInput
 
 
 def nnpinput_flatten(nnpinput: NNPInput):
--- a/modelforge/jax.py
+++ b/modelforge/jax.py
@@ -42,12 +42,17 @@ def convert_NNPInput_to_jax(nnp_input: NNPInput):
     nnp_input.box_vectors = convert_to_jax(nnp_input.box_vectors)
     nnp_input.is_periodic = convert_to_jax(nnp_input.is_periodic)
 
-    if nnp_input.pair_list is not None:
-        nnp_input.pair_list = convert_to_jax(nnp_input.pair_list)
+    nnp_input.pair_list = convert_to_jax(nnp_input.pair_list)
+    nnp_input.per_atom_partial_charge = convert_to_jax(
+        nnp_input.per_atom_partial_charge
+    )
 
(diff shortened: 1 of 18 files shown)


modelforge/jax.py
```diff
@@ -42,12 +42,17 @@ def convert_NNPInput_to_jax(nnp_input: NNPInput):
...
-        nnp_input.per_atom_partial_charge = convert_to_jax(
-            nnp_input.per_atom_partial_charge
-        )
+    # if nnp_input.pair_list is not None:
+    #     nnp_input.pair_list = convert_to_jax(nnp_input.pair_list)
+    #
+    # if nnp_input.per_atom_partial_charge is not None:
+    #     nnp_input.per_atom_partial_charge = convert_to_jax(
+    #         nnp_input.per_atom_partial_charge
+    #     )
```
wiederm, line 56: might be good to remove this

modelforge/tests/conftest.py
```diff
@@ -252,7 +252,7 @@ def generate_uniform_quaternion(u=None):
     """
     Generates a uniform normalized quaternion.
 
-    Adapted from numpy implementation in openmm-tools
+    Adapted from numpy implementation in modelforgeopenmm-tools
```
wiederm, line 255: I think this is a search/replace/ mistkae

modelforge/tests/test_potentials.py
```diff
@@ -1107,3 +1107,37 @@ def test_loading_from_checkpoint_file():
...
+    "potential_name", _Implemented_NNPs.get_all_neural_network_names()
+)
+def test_saving_torchscript(potential_name, single_batch_with_batchsize, prep_temp_dir):
+    batch = single_batch_with_batchsize(
+        batch_size=1, dataset_name="QM9", local_cache_dir=str(prep_temp_dir)
+    )
+    from modelforge.potential.potential import NeuralNetworkPotentialFactory
+    from modelforge.utils.misc import load_configs_into_pydantic_models
+
+    config = load_configs_into_pydantic_models("ani2x", "qm9")
```
wiederm, line 1122: It's fine to hardcode the dataset (qm9), but I think the potential_name should be passed here from the pytest parameterized input signature (otherwise the parameter doesn't impact the test and it's only testing for ani2x)

modelforge/utils/prop.py
```diff
@@ -105,12 +103,14 @@ def to_device(self, device: torch.device):
...
-        if self.pair_list is not None:
-            self.pair_list = self.pair_list.to(device)
-
-        if self.per_atom_partial_charge is not None:
-            self.per_atom_partial_charge = self.per_atom_partial_charge.to(device)
+        # if not self.pair_list is None:
+        #     self.pair_list = self.pair_list.to(device)
+        #
+        # if not self.per_atom_partial_charge is None:
+        #     self.per_atom_partial_charge = self.per_atom_partial_charge.to(device)
```
wiederm, line 113: this can be removed now

wiederm (approved):
This are good changes! Can you please update the test to cover all implemented potentials?

modelforge/tests/test_potentials.py
chrisiacovella: Good catch.  This was just a typo I fixed locally but apparently never committed/pushed my change on. The intent was to check all potentials to ensure we can write and read...not just test ani2x 7 times.

Issue #297, opened 2024-10-23
Neighborlists in TorchScript models


chrisiacovella:
Currently, when we set up a Potential, we pass to the constructor the neighbor list.  That means if we wish to change the neighbor list strategy (e.g., brute vs. verlet or any parameters therein), we need to instantiate and save a new version of the potential.  

Since part of our current plan is to distribute models as .pt files, rather than instantiating a class and loading in parameters (as this would require having modelforge be a dependency), this would require a separate pt file for all variations we might have, which is probably less desirable. 

While there are numerous ways we could address this, but I think the least disruptive approach is to merge the inference neighbor lists into a single class.  The brute force neighbor list is simply just the verlet list with a rebuild time of 1 and skin = 0 (although can also just put in some logic statements to switch which routines).   Even a cell based backend is simply just a modification to how we initially build the set of possible pairs.  

We can have a default strategy and allow users to set the strategy at runtime (e.g., via the constructor of the wrapping class designed to work with openmm-torch).

PR #299, opened 2024-10-25
Openmm integration

chrisiacovella:
## Pull Request Summary
This PR revamps the neighbor list (allowing us to change the strategy for a JITTED torchscript  potential) and provide a wrapper allowing us to run this with openmm -torch. 

### Key changes
 - [x] Neighborlists  include "only_unique_pairs" as a buffer
 - [x] Neighborlists for inference combined into a single class, with the ability to set the strategy either at run time or via function call (allowing a single model file)
 - [x] wrapper that follows the structure of TorchForce for openmm
 - [x]  This creates a submodule for openmm  integration (via modelforge.openmm)
 - [x] Tests for modelforge-openmm wrapping/running openmm 

### Associated Issue(s)
 - [x] #297 

## Pull Request Checklist
 - [x] PR is ready for review


chrisiacovella:
Note this is currently failing due to reading in a checkpoint file;  I added the unique_pairs variable to the buffers in the neighbor list, as that seemed to be a very import thing to set.  However, it is also set up stream (and encoded likely in another location).  So I wonder if this is necessary.

chrisiacovella:
The switch from a named tuple to a class for NNPInput actually presents some problems when we go to use the torch script models (i.e., passing input after calling torch.jit.script).   Torch script doesn't really handle getting a class as input well.  it sees the the NNPInput class as a different type due to being different Python compilation unit (i.e., compiled using modelforge and then compiled by the wrapper class).  

NNPInput is more of a convenience thing (has validation, sets default shapes, etc.), but for inference we just need to pass tensors directly.

wiederm:
@chrisiacovella , NNPInput is torch scripted right now; we could also set a flag not to jit that class (there is no benefit to it at that point).
An alternative is to pass input as keyword arguments, which should also be fine and need minimal changes.

modelforge/potential/neighbors.py
```diff
@@ -704,7 +707,8 @@ def forward(self, data: NNPInput):
 
         else:
             pairs_full, d_ij_full, r_ij_full = self._copy_to_nonunique(
-                self.nlist_pairs[:, in_cutoff],
+                self.nlist_i_pairs[in_cutoff],
```
wiederm, line 710: I like the changes here and in l699ff, this is much more readable

modelforge/potential/potential.py
```diff
@@ -305,6 +305,63 @@ def _remove_pairlist(self, processed_output: Dict[str, torch.Tensor]):
...
+        """
+        self.neighborlist._set_strategy(strategy, skin=skin)
+        # self.neighborlist.strategy = strategy
+        # self.neighborlist.skin = skin
+        # self.neighborlist.skin = skin
+        # self.neighborlist.half_skin = skin * 0.5
+        # self.neighborlist.cutoff_plus_skin = self.neighborlist.cutoff + skin
+
+    @torch.jit.export
+    def forward_for_jit_inference(
```
wiederm, line 329: can you clarify what this function does?

wiederm (commented):
Great work!

I'd love to better understand the control flow from input to property with this module, I left a comment.

modelforge/potential/potential.py
chrisiacovella: I will add some comments to the code and add it to the docstrings as well.  Basically, I just created a separate forward function because I couldn't pass the NNPInput class directly (as discussed above, it sees NNPInput as a different class because it was compiled outside...basically torch script does not like accepting class functions; it wants tensors or basic types like dicts of tensors).  This function allows us to just pass the individual values when calling from the openmm wrapper.   This was the simple work around because I think NNPInput is very useful and should still be used.

wiederm: Thank you for clarfiying! That's a good rational!

wiederm (approved):
Great work!

Issue #298, opened 2024-10-25
Force unit in datasets


wiederm:
I might misunderstand the curation pipeline, but it seems to me that energies and positions are correctly in kJ/mol and nanometer, but the force in some of the [datasets](https://github.com/choderalab/modelforge/blob/e844669739d573ddd80948f79b9c6b96385ec11e/modelforge/curation/phalkethoh_curation.py#L68) is in kJ/mol/angstrom. 

@chrisiacovella , can you clarify if I am missing something here?

chrisiacovella:
This seems to have been a mistake (or an accidental hold over when we actually were using angstroms internally).  

I think this is a 2 stage fix:

1. Add unit conversion into the data loader.  We have discussed this in the past as an extra measure to ensure all units are consistent, but we had not prioritized it.   this will allow us to use existing datasets without our target forces being an order of magnitude smaller. 
2. Change units in the datasets.  This is easy, but uploading to zenodo is a bit time consuming. As discussed in a separate issue #267, we can simplify our additional tags in the dataset (that tell us how to parse the numpy arrays) to only being  per_system, per_atom, and metadata.  changing units, these optional tags, and associated loader, can all be done in the same PR. 


I will do a PR for stage 1 now.

chrisiacovella:
#1 has been address in PR #301.   As mentioned in the PR, it would be good to now include a test to ensure proper unit conversion.  This will require creating a new test dataset, so I'll do that in a separate PR.

PR #301, opened 2024-10-25
refactoring HDF5Dataset class to check and convert units.

chrisiacovella:
## Pull Request Summary
This address the issue related to the force units being off (per angstrom, not per nanometer) in the datasets. 

### Key changes
 - [x] Units are attached to data read in and converted to our target system
 - [x] Each dataset now includes a mapping of a given property in the dataset to one of our available data fields (E, F, dipole_moment, total_charge). 
 - [x]  I commented out any available property that doesn't work in our current system (e.g., per_atom_charges, vibrational frequency, etc.), as they would be read in but couldn't be used in any context

### Questions
 - [x] While we dont' necessarily intend to embed using per-atom charges for our applications, do we want that to be an option? this might require more extensive discussion, just listing it because I've removed our ability to read any of those properties in at the moment. 

### Associated Issue(s)
 - #298 

## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/modelforge/utils/prop.py
+++ b/modelforge/utils/prop.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 import torch
 from typing import NamedTuple, Optional
 from loguru import logger as log
+from openff.units import unit
 
 
 @dataclass
@@ -16,6 +17,16 @@ class PropertyNames:
     F: Optional[str] = None
     total_charge: Optional[str] = None
     dipole_moment: Optional[str] = None
+
+
+PropertyUnits = {
+    "atomic_numbers": "dimensionless",
+    "positions": unit.nanometer,
+    "E": unit.kilojoule_per_mole,
+    "F": unit.kilojoule_per_mole / unit.nanometer,
+    "total_charge": unit.elementary_charge,
+    "dipole_moment": unit.elementary_charge * unit.nanometer,
+}
 
 
 class SpeciesEnergies(NamedTuple):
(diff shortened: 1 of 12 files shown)


modelforge/utils/prop.py
```diff
@@ -18,6 +19,16 @@ class PropertyNames:
...
 
+PropertyUnits = {
+    "atomic_numbers": "dimensionless",
+    "positions": unit.nanometer,
+    "E": unit.kilojoule_per_mole,
+    "F": unit.kilojoule_per_mole / unit.nanometer,
+    "total_charge": unit.elementary_charge,
+    "dipole_moment": unit.elementary_charge * unit.nanometer,
+}
+
```
wiederm, line 30: that's really useful!

wiederm (approved):
Looks great! Thanks for the quick fix!

wiederm:
> While we dont' necessarily intend to embed using per-atom charges for our applications, do we want that to be an option? this might require more extensive discussion, just listing it because I've removed our ability to read any of those properties in at the moment.

I think that's fine. I can't imagine a scenario in which partial charges as input features might be relevant. Even if we were to train on a particular partial charge scheme, we would still do it using element-specific electron negativity as input (which is fixed and can be put inside the model).

wiederm:
It might be good to add a test that checks that the first datapoint in each dataset has the correct implied units for energies and forces

modelforge/utils/prop.py
chrisiacovella: YES, I think so too!  It should help make things clearer for anyone looking at the code.

chrisiacovella:
> It might be good to add a test that checks that the first datapoint in each dataset has the correct implied units for energies and forces

This is a good idea.  I'll need to create a new, fake dataset to do this sort of testing.  I can do this in a separate PR (especially since this is merged).

PR #304, opened 2024-10-29
Fix center of mass shifting

chrisiacovella:
Center of mass shifting was failing when a dataset had multiple conformers (accidentally was using indices for the atom_numbers, which are not duplicated between multiple conformers, instead of using the indices that referred to series data).  As such, this was working fine for qm9, which was used for testing, but failing for all other datasets PhAlkEthOH. 

### Key changes
 - [x] Revise indexing in center of mass removal
 - [x] Add tests over PhAlkEthOH that also include ensuring r_ij and d_ij match (i.e., ensure we are invariant under translation)

### Associated Issue(s)
 - [ ] Issue 1

## Pull Request Checklist
 - [ ] Issue(s) raised/addressed and linked
 - [ ] Includes appropriate unit test(s)
 - [ ] Appropriate docstring(s) added/updated
 - [ ] PR is ready for review


wiederm (approved):
Thanks! That should fix the issue!

Issue #326, opened 2024-12-02
Tag name length over the limit of WandBLogger maximum


MarshallYan:
The _generate_tag function https://github.com/choderalab/modelforge/blob/f8c783461e5e2433b67185a938128cbc58656f14/modelforge/train/training.py#L1898-L1910 generates a tag that might exceed the length limit of what WandBLogger allows when multiple loss terms are included. We could use abbreviates (e.g. "ps" instead of "per_system" etc.) in case tags get too long.

Issue #325, opened 2024-11-26
plotting causing core dump when using profiler


chrisiacovella:
Both @MarshallYan  and I have encountered instances where training crashes due to issue with tkinter library (which is the python interface to tcl/tk).  This library is used under the hood by things like matplotlib, called during the plot generation. 

One of the errors is below: 

`Exception ignored in: <function Variable.__del__ at 0x7fb362735300>
Traceback (most recent call last):
  File "/home/shuaiy/miniforge3/envs/modelforge/lib/python3.11/tkinter/__init__.py", line 410, in __del__
    if self._tk.getboolean(self._tk.call("info", "exists", self._name)):
RuntimeError: main thread is not in main loop

Exception ignored in: <function Variable.__del__ at 0x7fb362735300>
Traceback (most recent call last):
  File "/home/shuaiy/miniforge3/envs/modelforge/lib/python3.11/tkinter/__init__.py", line 410, in __del__
    if self._tk.getboolean(self._tk.call("info", "exists", self._name)):
`

This only occurs when we have profilers set.  This seems to happen at different "times" (i.e., epochs).  E.g., crashing on epoch 18 one time 48 the next, 97 a third time (when using the simple profiler) 

Since this only happens during profiling, which typically will only typically run a few steps (and it seems to happen randomly), I'm not sure how much effort we should put into trying to better diagnose this.  If it crashes too early to get data, one can reduce number of training steps or simple try running again.

chrisiacovella:
Additional note, this occurs when using tensor board locally on my machine (which takes in the matplotlib figures and handles things after that...it might not happen in wandb).  @MarshallYan tried setting the matplotlib backend to `agg` and that solves the issue.  Will push a quick PR for this.

chrisiacovella:
`import matplotlib
matplotlib.use("Agg")
`

Added to the top of training.py

Issue #324, opened 2024-11-25
Dataset Toml files


chrisiacovella:
The dataset classes were set up to make it easy to load specific data keys from the hdf5 files and make it so different data keys can be set to the properties (e.g., assigning the data from "scf_dipole" to the property "dipole_moment" or "total_energy" to "E").  

However, since adopting the .toml file approach, we lost the ability to toggle this.  This functionality is important as we begin to test some subtle differences (e.g., scaling or not scaling the computed dipole moment). 

I think the fix can be rather simple, which would just be to add additional pieces to the pydantic model.  Basically we need two other sections:

- `dataset_keys` -- which keys to load from the dataset
- `properties_assignment` -- which keys to assign to each property (E, F, total_charge, etc.)

While this should be pretty straightforward, we'll want to design things such that we have a lot of validation in place (we'll need a list in each dataset file that specifies which data_keys are available).  This also links to another thing I want to address which is summarizing the datasets better.  for each dataset we should provide a list of available properties to load, their units, etc.

PR #327, opened 2024-12-03
Improved dataset control

chrisiacovella:
## Pull Request Summary
This provides changes aimed at improving the control over datasets.  Specifically, the switch to .toml files meant we could not easily toggle which properties to load from the hdf5 files nor which properties to assign to the key internal variables (energy, force, dipole_moment, etc.). 

This updates the pydantic dataset model whereby users now additionally define properties of interest (as a list) and then the association of these properties to the internal variable names (i.e used by the `Properties` class).  The pydantic model validates that anything in the Properties association also exists in the properties_of_interest list.   We do not validate that the properties_of_interest are actually in the hdf5 file via the pydantic model; this is handled when we assign those parameters while calling the `DataModule` class.  Since this is one of the earliest steps in the process, failure to properly define these should lead to a rapid failure.  this saves considerable extra code in the pydantic model, and having to make separate models for each dataset. 

 
### Key changes
 - [x] Allow better control over datasets via .toml files
 - [x] Fix buffer issue when logging to wandb
 - [x] Fix logic whereby using dipole moment in the loss required total charge as well (since dipole moment requires charges.  This is fixed such that charges are calculated if dipole moment is requested, even if total charge is not in the loss). 
 - [x] bug fix to spice1_openff curation (wrong tag for scf_dipole). 
 - [x] change backend for matplotlib to prevent issues when using debugger

### Associated Issue(s)
 - [ ] #326 
 - [ ] #325 
 - [ ] #324 

## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/modelforge/dataset/parameters.py
+++ b/modelforge/dataset/parameters.py
@@ -1,0 +1,100 @@
+from enum import Enum
+from typing import Optional, List, Dict
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+from typing_extensions import Self
+
+
+class CaseInsensitiveEnum(str, Enum):
+    """
+    Enum class that allows case-insensitive comparison of its members.
+    """
+
+    @classmethod
+    def _missing_(cls, value):
+        for member in cls:
+            if member.value.lower() == value.lower():
+                return member
+        return super()._missing_(value)
+
+
+# To avoid having to set config parameters for each class,
+# we will just create a parent class for all the parameters classes.
+class ParametersBase(BaseModel):
+    model_config = ConfigDict(
+        use_enum_values=True,
+        arbitrary_types_allowed=True,
+        validate_assignment=True,
+        extra="forbid",
+    )
+
+
+class DataSetName(CaseInsensitiveEnum):
+    QM9 = "QM9"
+    ANI1X = "ANI1X"
+    ANI2X = "ANI2X"
+    SPICE1 = "SPICE1"
+    SPICE2 = "SPICE2"
+    SPICE1_OPENFF = "SPICE1_OPENFF"
+    PHALKETHOH = "PhAlkEthOH"
+    TMQM = "tmQM"
+
+
+class PropertiesDefinition(ParametersBase):
+    atomic_numbers: str
+    positions: str
+    E: str
+    F: Optional[str] = None
+    dipole_moment: Optional[str] = None
+    total_charge: Optional[str] = None
+
+
+class DatasetParameters(BaseModel):
+    """
+    Class to hold the dataset parameters.
+
+    Attributes
+    ----------
+    dataset_name : DataSetName
+        The name of the dataset.
+    version_select : str
+        The version of the dataset to use.
+    num_workers : int
+        The number of workers to use for the DataLoader.
+    pin_memory : bool
+        Whether to pin memory for the DataLoader.
+    regenerate_processed_cache : bool
+        Whether to regenerate the processed cache.
+    properties_of_interest : List[str]
+        The properties of interest to load from the hdf5 file.
+    properties_assignment : PropertiesDefinition
+        Association between the properties of interest and the internal naming convention
+    """
+
+    model_config = ConfigDict(
+        use_enum_values=True, arbitrary_types_allowed=True, validate_assignment=True
+    )
+
+    dataset_name: DataSetName
+    version_select: str
+    num_workers: int = Field(gt=0)
+    pin_memory: bool
+    regenerate_processed_cache: bool = False
+    properties_of_interest: List[str]
+    properties_assignment: PropertiesDefinition
+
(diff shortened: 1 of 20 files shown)


modelforge/dataset/parameters.py
```diff
@@ -0,0 +1,100 @@
+from enum import Enum
+from typing import Optional, List, Dict
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+from typing_extensions import Self
+
+
+class CaseInsensitiveEnum(str, Enum):
```
MarshallYan, line 7: What does enum do?

modelforge/dataset/parameters.py
```diff
@@ -0,0 +1,100 @@
...
+    Class to hold the dataset parameters.
+
+    Attributes
+    ----------
+    dataset_name : DataSetName
+        The name of the dataset.
+    version_select : str
+        The version of the dataset to use.
+    num_workers : int
+        The number of workers to use for the DataLoader.
```
MarshallYan, line 62: Probably specify num_workers means the number of cores or threads used for loading?

modelforge/tests/data/config.toml
```diff
@@ -33,6 +33,12 @@ dataset_name = "QM9"
 version_select = "nc_1000_v0"
 num_workers = 4
 pin_memory = true
+properties_of_interest = ["atomic_numbers", "geometry", "internal_energy_at_0K", "dipole_moment"]
```
MarshallYan, line 36: We may need documentation for available options and possibly corresponding types and shapes (e.g., a scalar, an array with a size equal to the number of atoms).

modelforge/tests/data/config.toml
MarshallYan, line 1: Essential for training with dipole moment.

modelforge/train/training.py
```diff
@@ -45,6 +45,10 @@
 from modelforge.train.losses import LossFactory, create_error_metrics
 from modelforge.train.parameters import RuntimeParameters, TrainingParameters
 
+import matplotlib
+
+matplotlib.use("Agg")
```
MarshallYan, line 50: Need a comment here saying this is to get rid of tkinter errors in profilers by using background mode of matplotlib without calling GUI functions.

modelforge/train/training.py
```diff
@@ -1899,14 +1911,27 @@ def _generate_tags(self, tags: List[str]) -> List[str]:
         """Generates tags for the experiment."""
         import modelforge
 
+        try:
+            version = modelforge.__version__
+        except:
+            # for editable local install
```
MarshallYan, line 1917: Probably specify this is for installation with the "-e" flag.

modelforge/train/training.py
```diff
@@ -147,6 +151,10 @@ def __init__(self, requested_properties: List[str]):
         self.requested_properties = requested_properties
         self.include_force = "per_atom_force" in self.requested_properties
         self.include_charges = "per_system_total_charge" in self.requested_properties
+        # dipole_moment is calculated from charges, thus if dipole moment is requested
+        # we also need to calculate charges, even if we don't call per_system_total_charge
+        if "per_system_dipole_moment" in self.requested_properties:
```
MarshallYan, line 156: Need documentation with the dependency of properties, probably together with the list of properties available.

MarshallYan (approved):

modelforge/dataset/parameters.py
chrisiacovella: I defined `CaseInsensitiveEnum` to save effort later ( I added in a method that allows us to ignore string capitalization during assignment), and then can just subclass uses this as the parent .  More broadly speaking, setting up a class that inherits from Enum is the way to specify "choices" in a pydantic model, basically limiting what values can be accepted as input (it will produced a validation error if you give something not listed in the enum).

modelforge/dataset/parameters.py
chrisiacovella: I think this is most accurately the number of threads that are launched.  I'll note that testing a range of values did not produce too much of a difference (made more of a difference for the pair list precalculations which use the data loader class but still maybe 10-20%).  With the code more stable now, might be worth doing some short benchmarks.  As we discussed today, this might make more of a difference if datasets are much larger, but we also risk increasing memory usage if they are too large.

modelforge/tests/data/config.toml
chrisiacovella: this is a good point.  We need to have in the documentation and wiki summarizes of each dataset and what the available properties of interest are.  I can work on that in a separate PR (likely best to set up a script to autogenerate this .rst file)

modelforge/train/training.py
chrisiacovella: Same as comment above; we need to have this in documentation. Good idea.

Issue #321, opened 2024-11-20
Dataset API


chrisiacovella:
In discussions with @MarshallYan, the idea of revamping the curation to make it easier to add in a dataset.  While parsing out other file formats is going to need to be an ad hoc operation, we can probably ditch the base class in favor of an API to populate a pydantic class. 

To do this we could likely have a few "base" pydantic classes. In general, every entry is going require the following info.   

```
class DatasetProperty(BaseModel):
     name: str
     value: numpy.array
     unit: unit.Quantity
     value_shape_definition: Enum(per_atom, per_system) <-- we can either infer from class type or manually set
```

However, we likely would want different validation for each dataset property (per_atom vs per_system).  We could also likely subclass very specific cases, such as: "geometry", which would have shape (n_configs, n_atoms, 3) and units compatible with distance (both with we can validate); "energy", "forces", "atomic_numbers", "charges", "total_charge", etc. all which we can do the same validation.  This could help avoid someone giving the wrong dimensions (especially easy if dealing with a system with only a single config). 

We still can use the generic per_atom and per_system classes for types that don't fit within the defined ones.  Fields in a Metadata class wouldn't require unit and would need flexible value types (str, float, int, numpy.array).  The classes themselves (whether the inherit from per_atom or per_system, or if we just set value_shape_definition in each type of class) will also contain the info about what how to tag this data for ease of reading in, without needing to set it in an array like is currently done. 

In terms of assembling a system, we could simply add a property to a dictionary (I think wrap this in a class to make the interface easier). E.g., a class like

```
class DataSet:
     def __init__(self, dataset_name:str):
             self.dataset_name = dataset_name
             self.records = {}

       # add a record to the Dataset by giving it a unique name
      def register_record(record_name:str)
          # add some code to see if a record already exists and alert if duplicate
          self.records[record_name] = {}

      # add a property to the record, just a wrapper to make it easier to access underlying dictionary
      def register_property(record_name:str, property:DatasetProperty):
          self.records[record_name][property.name] = property

     def save_hdf5_file(filename:str):
          # call a validate function that will loop over Properties in each record checking that things like 
          # number of configs is consistent across all entries  
          self.validate_input()

          # we want to convert units to our target output
           self.convert_units()

          # call the hdf5 saving function. 
          self.save_hdf5(filename)

      # add some accessor functions to grab individual properties as well to then manipulate

```

[...]

PR #334, opened 2025-01-17
Dataset api

chrisiacovella:
## Pull Request Summary
This adds in a new module called "curate" that provides an API for dataset curation.  

The general hierarchy is that each dataset contains records, and each record contains properties.  In this, properties are defined using pydantic models (e.g., `AtomicNumbers`, `Energy`, `Forces`, `PartialCharge`, etc.) .  The pydantic property models ensure that for each value we ensure: units, property type (e.g., length, energy, force), and classification (per_atom, per_system), and also validates that the shape of the inputted array matches the expectation related to the classification.  Records collect the properties and validate that consistent number of configurations and atoms exist across properties.  The dataset class provides functions to write to an hdf5 file, converting units to the specified unit system. 

The dataset also can be initialize in an "append" mode whereby properties associated with individual configurations can be added separately, and the code will automatically append to the internal numpy array (performing any unit conversion necessary). 

This PR is still a WIP, as it will require also converting old curation scripts to use the new api, and to revise the dataset class in model forget to accept the new format (minor changes to the terminology we use for classification). 

### Key changes
 - [ ] Change 1

### Questions
 - [ ] Question 1

### Associated Issue(s)
 - [ ] #321 

## Pull Request Checklist
 - [ ] Issue(s) raised/addressed and linked
 - [ ] Includes appropriate unit test(s)
 - [ ] Appropriate docstring(s) added/updated
 - [ ] PR is ready for review


chrisiacovella:
The NPZ file that is generated from the dataset is somewhat dynamic (and we can't just look at the md5 checksum).  To see if the code needs to regenerate that npz file or just use the one already that exists, we look at a metadata file that is written out (as a json file). This include the md5checksum of the hdf5 file and also the properties of interest that were used to generate the npz file.  We need to also put the element filter in there and look at it.  If the element filter has changed, we need to regenerate that npz file.

Issue #345, opened 2025-03-21
Fix bug in tmqm_xtb dataset


chrisiacovella:
tmqm_xtb dataset had a small bug in determining which configurations to include (used wrong indexing).  The functions that perform the filtering also could be more general, i.e. accessible outside of the class that actually curates the dataset, as they would be useful for examining the datasets in a Jupyter notebook.

Issue #344, opened 2025-03-21
Make it easier to work with HDF5 files


chrisiacovella:
As the title suggests, we can make it easier to work with and interrogate the HDF5 files we are generating.

Some things I've implemented:
- Read an HDF5 file we generated back into a SourceDataset instance
- Add in function to automatically convert a Record to a RDKit molecule for easier visualization and analysis
- provide a function for inplace unit conversions of properties
- provide a function to convert all properties in a record to the global unit system
- provide a function to convert all properties in all records in a dataset to the global unit system

PR #346, opened 2025-03-21
Extras for dataset api to make it easier to work with HDF5 datafiles

chrisiacovella:
## Pull Request Summary
This adds in additional functionality to the Dataset API to make it easier to interrogate datasets and fixing a bug in the tmqm_xtb curation. 

### Key changes
 - [x] fix tmqm_xtb curation bug
 - [x] Read an HDF5 file we generated back into a SourceDataset instance
 - [x] Add in function to automatically convert a Record to a RDKit molecule for easier visualization and analysis
 - [x] provide a function for inplace unit conversions of properties
 - [x] provide a function to convert all properties in a record to the global unit system
 - [x] provide a function to convert all properties in all records in a dataset to the global unit system
 - [x] update Jupyter notebooks to show how to use these functions (partially done; need to show RDKit conversion/visualization)

### Associated Issue(s)
 - [x] #344 
 - [x] #345 

## Pull Request Checklist
 - [x] PR is ready for review


modelforge-curate/modelforge/curate/properties.py
```diff
@@ -181,6 +182,32 @@ def n_atoms(self) -> int:
...
+                )
+        elif self.classification != "atomic_numbers":
+            if not units.is_compatible_with(
+                GlobalUnitSystem.get_units(self.property_type), "chem"
+            ):
+                raise ValueError(
+                    f"Unit {units} of {self.name} are not compatible with the property type {self.property_type}.\n"
+                )
+
+            self.value = (self.value * self.units).to(units, "chem").m
```
MarshallYan, line 208: Same as the previous comment here.

What is `*contexts` and what are the options?

modelforge-curate/modelforge/curate/tests/test_curate.py
```diff
@@ -100,6 +100,23 @@ def test_dataset_create_record(prep_temp_dir):
         new_dataset.add_records([record6, record7])
 
 
+def test_convert_record_units():
```
MarshallYan, line 103: Should there be a negative test here (e.g. Wrong input units)?

modelforge-curate/modelforge/curate/tests/test_curate.py
```diff
@@ -193,6 +223,171 @@ def test_record_failures():
...
+    record.add_properties([positions, energies, atomic_numbers, meta_data])
+
+    record.remove_property("positions")
+    assert "positions" not in record.keys()
+    assert "positions" not in record.per_atom.keys()
+    record.remove_property("energies")
+    assert "energies" not in record.keys()
+    assert "energies" not in record.per_system.keys()
+
+    record.remove_property("smiles")
```
MarshallYan, line 293: This might not be implemented in this PR, but can we remove `meta_data.smiles` rather than "smiles" since it is a property under `meta_data`. Or what is the hierarchy of these properties?

modelforge-curate/modelforge/curate/properties.py
```diff
@@ -181,6 +182,32 @@ def n_atoms(self) -> int:
...
+        """
+
+        if self.classification == "meta_data":
+            if self.units != unit.dimensionless:
+                log.info(
+                    f"{self.name} has units {self.units}, but is MetaData and will not be converted."
+                )
+        elif self.classification != "atomic_numbers":
+            if not units.is_compatible_with(
+                GlobalUnitSystem.get_units(self.property_type), "chem"
```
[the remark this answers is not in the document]
chrisiacovella: openff.units is based off of the pint package.  This allows you to define specific contexts for unit conversion/compatibility.  In this case, "chem" is a context I've added that allows us to convert units of energy (like Hartree) to units of energy per mole (like KJ/mol).  See: https://github.com/choderalab/modelforge/blob/main/modelforge/utils/units.py

modelforge-curate/modelforge/curate/tests/test_curate.py
chrisiacovella: Good catch.  I did not have any validation when adding a Record that it was actually an instance of the Record class.  I added this and put in tests to check that passing something else will fail.

modelforge-curate/modelforge/curate/tests/test_curate.py
chrisiacovella: The name of a property is unique.  If we have "smiles" in meta_data, it can't also be in "per_atom".  I added the "remove_property" function just to make it easier for a user to not have to really understand the underlying hierarchy as much, since that underlying hierarchy (i.e., separate dictionaries for per_atom, per_system, and meta_data) really is just for book keeping and actually could probably be removed.  I might actually do that in a different PR...it would definitely remove a bunch of extra logic in the code.

MarshallYan (approved):

Issue #349, opened 2025-04-01
Revise dataset.py structure


chrisiacovella:
Currently, each of the supported datasets are defined in their own class that inherits from the HDF5Dataset class.  Each of these classes provides information about each dataset:
- properties available within the dataset
- default properties of interest
- properties of interest (to actually use)
- properties association (which specifies the type of each quantity, used for unit conversion)
- atomic self energies 

Previously, these also included the links to the files to download (and checksums) for each version, however these were moved into yaml files (see PR #117).  

To allow control over the properties used in training, definition of properties of interest and the association of individual keys in the dataset with internal variables (i.e., E: total_energy, positions: geometry, etc.), pydantic models were added that require these to be defined in the .toml runtime files (See PR #327 ); this makes much of what is currently in the dataset specific classes some what unnecessary. 

Plan: 

A better, more flexible approach would be to move the atomic self energies and  the available properties to the yaml files.  Initialization of the HDF5Dataset class would simply accept which dataset to use.  

The new dataset API now includes information about each property needed to validate units (i.e., ensuring that an energy property has the appropriate energy units we expect). I think moving the GlobalUnitSystem class defined in the curate module to the main repo and utilizing this during dataset loading would remove some of the hackiness currently in the code.  I also think that, the meat of the HDF5Dataset class (responsible for reading/parsing the dataset) could largely be replaced with the Dataset API. Basically reading the HDF5 file back into an instance of SourceDataset, allow the use of the dataset api for validation of the inputs.  that would also make it easier to perform future manipulation/subsetting of the dataset (see issues #306 and #329 ) as there is lots of functionality built into source dataset. 

Note: this would eliminate the current backwards compatibility with the older datasets, so it just requires that I also get all the newly curated datafiles available on zenodo at the same time. 

Additional, two other changes need to be made at the same time:
- change machine strategy such that we have a global_cache (e.g., for datasets) and a local_cache (for the output for a given run). 
- allow users to input use a local dataset (via a user specified yaml file). I think this will be important for testing new datasets and give the option to use something that is not available publicly.

chrisiacovella:
Regarding revising the .yaml files... 

To allow us to have a single hdf5 data loader, rather than individual python classes that inherit from the HDF5Dataset, we need to put more information in the .yaml files. 

[...]

PR #358, opened 2025-05-27
Dataset.py refactoring

chrisiacovella:
## Pull Request Summary
The general idea here is to refactor the way we handle datasets such that we no longer are required to create a unique child class of the HDF5Dataset class for each dataset.   Rather than a unique class for each dataset, the code now relies upon yaml files that contain the key information (available properties, atomic self energies, etc.).  These yaml files still contain the various versions of the dataset and link/doi to fetch the dataset.     Note, this switch removes default properties of interest, which is keeping with the philosophy of requiring users to define all inputs. (note, this however, broke all the tests that relied on this implicit information). 

We still maintain a list of "built-in" datasets, but this structure now allows us to support "local" datasets (i.e., those we haven't uploaded to zenodo).  The yaml file allows a local_datset to be defined, with the same basic syntax, just you provide a path to the dataset on your local system.  

This additionally drops support for the older hdf5 datafiles, now switching to those generated using the new dataset.curate module (i.e., that shifts validation to the time of creation of the dataset).  These datasets provide a bit more information (specifically the property_type, i.e., length, energy, force, etc.) that is useful for performing any unit conversion that is necessary when loading.   All supported datasets have been recurated and uploaded to zenodo, and yaml files updated to support new format. 

To this end, the global unit system class has been moved from curate to the main module of modelforge to streamline the unit conversion/checking when loading in the datafile. 

 
### Key changes
 - [x]  Set up curate scripts for a few additional datasets (ani1x, spice1 openff, spice2 openff)
 - [x] regenerate and upload all new versions of the datasets to zenodo 
 - [x] switch all public facing calls that require units to using the global units (e.g., so a cutoff is converted to whatever unit is in the global units.  this will allow us to presumably experiment with different unit systems and their impact on convergence)
 - [x] add in helper functions that call the dataset .toml files (in tests/data/datasets) to define default properties of interests and properties associations for testing (previously these defaults were just part of the dataset class).   This also makes it such that fewer pieces of code need to be updated if we want to test with a different version. 
 - [x]  Update all the tests to work with the new scheme
 - [x] Add in a class to make it easier to generate the version information that goes into a yaml file. 

### Associated Issue(s)
 - [x] #349 

## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/modelforge/dataset/dataset.py
+++ b/modelforge/dataset/dataset.py
@@ -308,6 +308,7 @@ class HDF5Dataset:
         if self.local_yaml_file is not None:
             yaml_file = self.local_yaml_file
             # make sure the file exists
+            # make sure the file exists
             if not os.path.exists(yaml_file):
                 raise FileNotFoundError(
                     f"Local dataset yaml file {yaml_file} not found."
@@ -1250,6 +1251,7 @@ class DataModule(pl.LightningDataModule):
         training_dataset_statistics = calculate_mean_and_variance(torch_dataset)
 
         # wrap everything in a dictionary and save it to disk
+
         dataset_statistic = {
             "atomic_self_energies": atomic_self_energies.energies,
             "training_dataset_statistics": training_dataset_statistics,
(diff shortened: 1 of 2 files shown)


modelforge/custom_types.py
MarshallYan, line 1: Are all datasets named with capital letters now?

MarshallYan (approved):

modelforge/custom_types.py
chrisiacovella: we use all lowercase for internal file names (and covert to all lower case internally when calling the files), so it shouldn't matter. When defining a name in the .toml files, we use a case insensitive enum to assign the name, so capitalization shouldn't matter.

PR #407, opened 2026-03-19
Feature nvalchemi d3 support

### LINKED ISSUE #402: NVALCHEMI toolkit support
- DFT-D3
Currently, within the code, we use TAD-DFTD3 for the VDW contributions.   While this package works fine, it will recalculate neighbor lists internally during the computation.  Computing this a second time obviously takes additional time which could further slow things down when in production.  Additionally, I'm not quite sure how efficient this neighbor listing is (I assume just brute force) which will further slow things down (e.g., not being able to take advantage of the verlet neighbor listing).  However, I'll note that one advantage of this package is that it is on condo-forge. 

https://github.com/NVIDIA/nvalchemi-toolkit-ops/tree/main
The nvalchemi toolkit provides a DFTD3 implementation (which is essentially just the implementation from aimnet2, cleaned up a bit).  This package takes as input the interacting pairs from the neighbor list.  As such, it will avoid additional recompilation.    I'll note that for periodic systems, it will require the shifts, which is not something our current neighbor list provides (periodicity and, in theory, different shaped boxes, are handled via a "space" class).   It should be straight forward to have this class return the shifts as well. 
I'll note one down side is that nvvalchemi is not on conda-forge at the current time (something we should look into). 

I'll note, we can easily support both methods; lazy-imports and the delayed import class I added (which will give warnings and tell you where to get the package) should be useful in this context.  Aside from the recomputing of neighbor lists, nvalchemi is billed as being "fast" but I have not yet tested that (documentation is poor and provides examples that do not actually work and are missing libraries/data). 

- Neighborlists
I'll note that the neighbor lists in modelforge do not necessarily return the interacting pairs in numerical order (where we could just count up the number of repeats in the first dimension of the tensor to get the number of neighbors... e.g., it could be[ 0, 0, 0, 1, 1, 2, 2, 1, 0, 2]).  This is because to calculate the "full" neighbor list, I found it was faster to just simply copy the half neighbor list (applying the appropriate  sign change). I think it would make sense to simply sort based on the first index in the [2, npairs] tensor.  This would be more general anyway.   

Regarding neighbor lists, I think I need to shift around some of the logic: while for local interactions the "unique_pairs_only" is an important toggle, when it comes to VDW the routines will require the full neighbor list regardless; electrostatics (in current implementation, require half, and it has a function in there to remove non-unique pairs).  We should just switch electrostatics to always calculate half, rather than full and then. making it half again;  this was certainly an oversight in terms of efficiency. 

I'll note, the toolkit does provide some of their own internal neighbor lists (specifically a cell list based one).  However, I'll note that these operate as traditional neighbor lists, simply providing pairs, rather than also returning r_ij and d_ij as well, as our code expected (this was a specific design choice which has some advantages). However, the design of the current neighbor lists does allow us to relatively easily swap in different backends (e.g., previously setting up versions that use nnp_ops).  This is something worth looking into; my previous testing (of other implementations) didn't really show much benefit for a cell-list based scheme as compared to a nsq^2-verlet until system size was really large (the overhead of dealing with cells canceled out the cost savings given the ability of current GPUs to handle a lot of simple calculations during rebuild). 

- Electrostatics
The nvalchemi toolkit also has various electrostatic methods (including Ewald summation, which I'd prefer not to have to figure out how to implement in pytorch).   This should again be straightforward to implement once other changes to the neighbor list are accomplished. 

I'll make sub issues for the various parts.

### LINKED ISSUE #403: Modify neighbor list
In the context of being able to work with the NVALCHEMI toolkit and simply to provide some speed improvements (and just more consistent behavior overall), the following refactoring will be needed: 

- the `unique_pairs_only` flag should apply only to local interactions. The VDW should always be the full and electrostatics should be half (at least for current implementation; need to check on nvalchemi).  A small refactor such that the function that handles this takes as input unique_pairs_only, where only the call for local neighbor list would use this. 
- When creating the full neighbor list, sort indices in the tensor so that indices are always sequential (for the first index in the interacting pairs tensor). 
- modify the space class to return shifts; shifts can be included in the `PairListOutput` dataclass as well.  Since right now we are mostly dealing with non-periodic systems, this is less urgent but should be done.

### LINKED ISSUE #404: Support for other packages for VDW
In addition to TAD-DFTD3, support the nvalchemi-toolkit-ops version. 
- Small refactoring in the code to accept different "schemes" (or library) in the .toml file and make this toggle happen in the code. 
- TAD-DFTD3 provided some utilities to batch the inputs (to be in the form they expect).  NVALCHEMI should be able to handle batched data in the same way we are exporting from our neighbor list, but this needs to be investigated more thoroughly to ensure calculations are correct and implemented in an efficient way.

MarshallYan:
## Pull Request Summary
Provide a brief description of the PR's purpose here.

### Key changes
Implemented another option (nvalchemiops) for dftd3 dispersion correction calculation.

### Associated Issue(s)
 - [x] Issue #402 
 - [x] Issue #403
 - [x] Issue #404

## Pull Request Checklist
 - [x]  Do a test training run 
 - [x] PR is ready for review


modelforge/potential/potential.py
```diff
@@ -431,13 +437,10 @@ def _add_pairlist(
...
         core_output["local_r_ij"] = pairlist_output.local_cutoff.r_ij
 
-        # this is commented out for now, as the vdw energy is calculated via
-        # DFTD3 which handles pair calculations internally
-        # If we have other vdw implementations, we can uncomment this
-        # if "per_system_vdw_energy" in self.postprocessing._registered_properties:
-        #     core_output["vdw_pair_indices"] = pairlist_output.vdw_cutoff.pair_indices
-        #     core_output["vdw_d_ij"] = pairlist_output.vdw_cutoff.d_ij
-        #     core_output["vdw_r_ij"] = pairlist_output.vdw_cutoff.r_ij
+        if "per_system_vdw_energy" in self.postprocessing._registered_properties:
```
chrisiacovella: This is fine for now...but tad-dftd3 doesn't make use of this information (so we may need to have some additional conditional statements here to check the backend being use). I think for training it won't make a difference, but in inference this might add some extra overhead if using tad-dftd3. 

I think add a comment to the code that this is only needed for nvalchemiops and in the future sure put some additional conditions on this.

chrisiacovella (approved):
A few minor, non-blocking comments. I think once these are addressed we can merge.  Reach out on slack if any of my comments aren't clear.

modelforge/potential/potential.py
```diff
@@ -329,12 +335,12 @@ def __init__(
         # note cannot jit compile the dispersion interactions as tad-dftd3 is not compatible with torchscript
         if "per_system_vdw_energy" in postprocessing._registered_properties:
             log.warning(
-                "JIT compiling the postprocessing module with vdw interactions will not work."
+                "JIT compiling the postprocessing module with vdw interactions will not work if using tad-dftd3."
             )
+            log.warning("tad-dftd3 packaged is not compatible with torchscript.")
             log.warning(
-                "The tad-DFTD3 packaged used is not compatible with torchscript."
+                "Use nvalchemiops as the DFTD3 engine or disable JIT compilation by setting jit=False."
```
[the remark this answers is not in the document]
MarshallYan: No, I haven't tested nvalchemiops with JIT. Should we just remove these logs?

modelforge/potential/potential.py
MarshallYan: I put a FIXME here since we don't have a usable postprocessing property that assign if tad-dftd3 or nvalchemiops is used.

Issue #414, opened 2026-04-19
Can't import the data module thus can't use the default d3 parameters outside the main project directory


MarshallYan:
Need to put __init__.py in the data directory.

MarshallYan:
When modelforge is install in editable mode (-e), modules in the package will be imported from the development directory. Since our d3 parameters are stored as a pt file under the development directory, training with nvalchemiops works. It will not work without the -e flag since the pt file is not copied to the installation directory (e.g., ./miniforge3/envs/modelforge/lib/python3.12/site-packages/modelforge/data/dftd3_parameters).

Not sure how this was handled in the dataset module where all the yaml files are in the installation as well.

PR #421, opened 2026-05-19
Init data module & send matrices to proper devices

MarshallYan:
…mi-ops)

## Pull Request Summary
Init modelforge/data as a module so that import can be done despite the running directory. Fix a typo of not copying a matrix to devices.

### Key changes
 - [x] Init modelforge/data as a python module so that its path can be imported
 - [x] Copy interact_fillers to proper devices (CPU or GPU), this was a typo from last nvalchemi-ops dispersion calculation

### Associated Issue(s)
 - [x] Issue #414 

## Pull Request Checklist
 - [x] PR is ready for review

### DIFF
--- a/modelforge/potential/processing.py
+++ b/modelforge/potential/processing.py
@@ -1095,7 +1095,9 @@ class DispersionPotential(torch.nn.Module):
         """
 
         # fill the matrix with ones at interacting pairs
-        interact_fillers = torch.ones(neighbor_list.shape[1], dtype=torch.int64, device=neighbor_list.device)
+        interact_fillers = torch.ones(
+            neighbor_list.shape[1], dtype=torch.int64, device=neighbor_list.device
+        )
 
         # construct a neighbor matrix where row & col indices indicates pairs of atoms, and interacting pairs are filled
         # with value one in the matrix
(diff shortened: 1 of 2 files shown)


modelforge/train/training.py
```diff
@@ -450,7 +450,7 @@ def _predict_quadrupole_moment(
             The predicted quadrupole moment for each system.
         """
 
-        """
+        r"""
```
chrisiacovella, line 453: Is this just a stray character?

modelforge/potential/processing.py
```diff
@@ -183,7 +181,6 @@ def __getitem__(self, key):
 
     def __iter__(self) -> Iterator[Dict[str, float]]:
         """Iterate over the energies dictionary."""
-        from modelforge.utils.units import chem_context
```
chrisiacovella, line 186: This needs to be imported somewhere in the file, otherwise we can't use the appropriate conversions of energy (well energy to energy/mol). 

this imports the context required for openff units that is called, e.g., on line 166

.to(GlobalUnitSystem.get_units("energy"), "chem")

chrisiacovella (changes requested):
Trying to figure out what was changed to address the nvalchemi import issue.

chrisiacovella:
Ah, I missed that you added in the _init__.py file.

modelforge/train/training.py
MarshallYan: The 'r' indicates that the doc string is a raw string since there are backward slashes in it and this might confuse the Python interpreter.

[Raw strings explained.](https://www.geeksforgeeks.org/python/python-raw-strings/)

